Issue #15, opened 2021-04-22
Error 400: Invalid value for field 'resource.sourceDisk'


danielkubat:
#### Overview of the Issue

When trying to use `googlecompute, v0.0.2` plugin, packer is failing to create an image after successful build phase. Using `packer 1.7.2` binary without `required_plugins` block and `init` works fine as before for very same code (template), in same environment.

#### Reproduction Steps

Execution of packer templates with defined `required_plugins` block after `packer init`

### Plugin and Packer version

`packer 1.7.2`
`plugin 0.0.2`

### Simplified Packer Buildfile
```
source "googlecompute" "default" {
  disk_size             = var.disk_size
  disk_type             = var.disk_type
  image_description     = var.desc
  image_family          = var.image_family
  image_labels          = var.image_labels
  image_name            = var.image_name
  labels                = var.labels
  machine_type          = var.machine_type
  network               = var.network
  omit_external_ip      = var.omit_external_ip
  preemptible           = var.preemptible
  project_id            = var.project_id
  scopes                = var.scopes
  service_account_email = var.service_account_email
  source_image_family   = var.source_image_family
  ssh_username          = var.ssh_username
  subnetwork            = var.subnetwork
  tags                  = var.tags
  use_internal_ip       = var.use_internal_ip
  zone                  = var.zone
}
```

### Operating system and Environment details

CentOS7/8, RHEL 7/8 Google provided images

### Log Fragments and crash.log files

```
Step #1: ==> googlecompute.default: Creating image...
Step #1: ==> googlecompute.default: Error waiting for image: googleapi: Error 400: Invalid value for field 'resource.sourceDisk': 'https://compute.googleapis.com/compute/v1/project/zones/europe-west3-a/disks/packer-60815dbe-bb2b-1e0b-fd89-fb85adceef25'. The URL is malformed., invalid
Step #1: ==> googlecompute.default: Deleting disk...
Step #1:     googlecompute.default: Disk has been deleted!
Step #1: ==> googlecompute.default: Provisioning step had errors: Running the cleanup provisioner, if present...
Step #1: Build 'googlecompute.default' errored after 5 minutes 45 seconds: Error waiting for image: googleapi: Error 400: Invalid value for field 'resource.sourceDisk': 'https://compute.googleapis.com/compute/v1/project/zones/europe-west3-a/disks/packer-60815dbe-bb2b-1e0b-fd89-fb85adceef25'. The URL is malformed., invalid
[...]

PR #16, opened 2021-04-23
Fix Error 400: Invalid value for field 'resource.sourceDisk

Filirom1:
Fix Error 400: Invalid value for field 'resource.sourceDisk' #15

### DIFF
--- a/builder/googlecompute/driver_gce.go
+++ b/builder/googlecompute/driver_gce.go
@@ -176,7 +176,7 @@ func (d *driverGCE) CreateImage(name, description, family, zone, disk string, im
 		Labels:             image_labels,
 		Licenses:           image_licenses,
 		ImageEncryptionKey: image_encryption_key,
-		SourceDisk:         fmt.Sprintf("%s%s/zones/%s/disks/%s", d.service.BasePath, d.projectId, zone, disk),
+		SourceDisk:         fmt.Sprintf("%sprojects/%s/zones/%s/disks/%s", d.service.BasePath, d.projectId, zone, disk),
 		SourceType:         "RAW",
 		StorageLocations:   imageStorageLocations,
 	}


danielkubat:
@SwampDragons @azr Can you please review this PR?

azr (approved):
Hello there, thanks for fixing it, LGTM.

brandonjbjelland:
Thanks a bunch for this fix! I'm hitting this same problem as I upgraded to HCL2 and migrated to this plugin. Great to see it repaired in no time.

Question for @azr or whoever owns release: Can you create a patch or minor release from the current contents of `main`? The `required_plugins` block requires that each plugin have a `version` specified so I gather anyone else using this plugin is currently stuck with packer runs that cannot finish and successfully publish images.

Many thanks for your work here! 🙏

danielkubat:
> Question for @azr or whoever owns release: Can you create a patch or minor release from the current contents of `main`? The `required_plugins` block requires that each plugin have a `version` specified so I gather anyone else using this plugin is currently stuck with packer runs that cannot finish and successfully publish images.

+1, we are waiting for new version release as well.

brandonjbjelland:
Also, somewhat odd that this bug came up at all. I understand the API has changed since moving to the separated plugin system but I would think this codebase would largely reflect the packer-embedded `googlecompute` code from pre Packer `1.7.x`. Mostly, I'm trying to understand if this plugin has taken any steps backward in maturity during the move and how to convey that to clients. If indeed the contents here are largely the same as before, I would suggest reflecting that in the version. Considering this builder has been around for essentially the lifetime of Packer, this is squarely feels like a `1.0.0` plugin or at the very least `0.1.0`. Thoughts?

SwampDragons:
> Mostly, I'm trying to understand if this plugin has taken any steps backward in maturity during the move and how to convey that to clients who would correctly question a pre 0.1.0 version of anything. 

Valid question. You're right that the content is very close to what it was when it was in the core, and that the functionality _should_ be the same. When first splitting the plugins out, we set them to a pre-1.0 value for a couple of reasons: 

1) We wanted to make it clear that there may be a couple of bugs that needed working out just because we had to do a lot of re-architecting to extract them from the Packer core, and if we had to make an early, backwards-incompatible change to a plugin to get the required_plugins block working or something, we wanted to have the flexibility to do that. This turned out to be true for a few plugins where the versioning package was messed up or the plugin name turned out to need changing in the very early releases and we needed to push quick iterative updates in the 0.0.x line. That said, I believe everything is pretty well settled now. 

2) We wanted to make it extremely clear that the plugin versioning is decoupled from the Packer core versioning. Starting at v1.7 might have caused confusion in that people may have expected plugins and core versions to stay in sync with each other.

Anyway, I think you can expect us to jump straight to a 1.0 release pretty soon, especially since like you said these plugins have been maturing for years in the Packer core repo.  I want to do one or two more patch releases of the packer _core_ in the 1.7.x line where all the plugins we wanted to extract have been vendored back in, and then I think I'll be comfortable promoting especially the bigger plugins like google and amazon to 1.0.

brandonjbjelland:
@SwampDragons - that all makes sense, thanks for the details and the plan feels reasonable. I'll keep an eye out for a patch release here.

brandonjbjelland:
And a release just landed today, testing it out now. Thanks for pushing that!

PR #22, opened 2021-05-25
Auto merge minor versions of github.com/hashicorp/packer-plugin-sdk

azr:
This is a test to see if this works. If this works I'll update other repos.


sylviamoss (approved):
I like this a lot! 🎉

PR #32, opened 2021-08-24
Add HCP Packer registry image metadata to builder artifacts

nywilken:
This change adds support for publishing HCP Packer registry images from existing builder artifacts with the Provider name set to "gce".  The images currently don't have any label information but would


builder/googlecompute/artifact.go
```diff
@@ -43,8 +45,14 @@ func (a *Artifact) String() string {
 }
 
 func (a *Artifact) State(name string) interface{} {
-	if _, ok := a.StateData[name]; ok {
-		return a.StateData[name]
+	if name == registryimage.ArtifactStateURI {
```
nywilken, line 48: There's some rich metadata that we can extract from the Artifact config here if set. I will revisit this tomorrow.

SwampDragons: Sounds good.

SwampDragons (approved):
Works as advertised. Used this plugin to run a gce build, upload it to PAR, and then was able to pull from PAR to use output from original build as a Packer source.

azr (approved):
LGTM ! 🙂

builder/googlecompute/artifact.go
nywilken: Labels added. Thanks again for the review.

PR #33, opened 2021-08-24
Use user-configured Service Account for public key import when running in GCE

### LINKED ISSUE #5: googlecompute builder ignores precedence of authentication methods when running in GCE
_This issue was originally opened by @rbjorklin as hashicorp/packer#10607. It was migrated here as a result of the [Packer plugin split](https://github.com/hashicorp/packer/issues/8610#issuecomment-770034737). The original body of the issue is below._

<hr>

#### Overview of the Issue

googlecompute builder ignores [precedence of authentication methods](https://www.packer.io/docs/builders/googlecompute#precedence-of-authentication-methods) when running in GCE. **The issue can be worked around by commenting [this line](https://github.com/hashicorp/packer/blob/fd8e76636a532d1f6eda68d1a2f718491ffbf2d4/builder/googlecompute/step_import_os_login_ssh_key.go#L44).**

#### Reproduction Steps

* Run a packer build with the ansible provisioner using `use_os_login` from your local workstation using `account_file`
* Try running the same build from within GCE itself and watch the build either fail or use other credentials than specified

### Packer version
Packer v1.7.0-dev

```hcl
source "googlecompute" "build-dev-agent" {
  service_account_email   = "<something>"
  image_name              = "jenkins-build-agent-${formatdate("YYYY-MM-DD't'hh-mm-ss", timestamp())}"
  image_family            = "jenkins-build-agent"
  use_os_login            = true
  project_id              = "<something>"
  source_image_family     = "centos-7"
  ssh_username            = "packer"
  zone                    = "<something>"
  network_project_id      = "<something>"
  subnetwork              = "<something>"
  omit_external_ip        = true
  use_internal_ip         = true
  account_file            = "sa_key.json"
}

build {
  sources = [
    "sources.googlecompute.build-dev-agent",
  ]

  provisioner "ansible" {
    host_alias      = "localhost"
    playbook_file   = "./playbook.yml"
  }
}
```

Centos 7, x86_64

```
packer-builder-googlecompute plugin: [INFO] Requesting Google token via account_file...
packer-builder-googlecompute plugin: [INFO]   -- Email: intended-account@intended-project.iam.gserviceaccount.com
...
packer-builder-googlecompute plugin: [INFO] OSLogin: GCE service account wrong-account@wrong-project.iam.gserviceaccount.com will be used for identity
==> googlecompute.build-dev-agent: Importing SSH public key for OSLogin...
==> googlecompute.build-dev-agent: Error importing SSH public key for OSLogin: googleapi: Error 403: End user credentials must match the user specified in the request., forbidden
```

jonsulman:
When running on a Compute Engine instance, `StepImportOSLoginSSHKey` uses the account name from the GCE instance even if a different service account is configured. This can cause Packer to authenticate to Google API using the configured service account but attempt to create an SSH key for the GCE service account, resulting in an error:

    googleapi: Error 403: End user credentials must match the user specified in the request

This changes the behavior and uses the GCE service account only if there is no service account in the configuration.

Closes #5

### DIFF
--- a/builder/googlecompute/step_import_os_login_ssh_key.go
+++ b/builder/googlecompute/step_import_os_login_ssh_key.go
@@ -20,6 +20,7 @@ type StepImportOSLoginSSHKey struct {
 	Debug         bool
 	TokeninfoFunc func(context.Context) (*oauth2.Tokeninfo, error)
 	accountEmail  string
+	GCEUserFunc   func() (string)
 }
 
 // Run executes the Packer build step that generates SSH key pairs.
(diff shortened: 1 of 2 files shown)


rickfoster101:
Running into this too. When can we expect a review and release? Thx.

builder/googlecompute/step_import_os_login_ssh_key.go
```diff
@@ -57,6 +51,17 @@ func (s *StepImportOSLoginSSHKey) Run(ctx context.Context, state multistep.State
 		s.accountEmail = config.account.jwt.Email
 	}
 
+	if s.GCEUserFunc == nil {
+		s.GCEUserFunc = getGCEUser
+	}
```
azr, line 56: 👋🏼 
Is there a case when `s.GCEUserFunc` is not nil here ? You just created the field and I don't see it be set anywhere else ?

builder/googlecompute/step_import_os_login_ssh_key.go
```diff
@@ -57,6 +51,17 @@ func (s *StepImportOSLoginSSHKey) Run(ctx context.Context, state multistep.State
...
+	if s.accountEmail == "" {
+		s.accountEmail = s.GCEUserFunc()
+	}
```
azr, line 59: 👋🏼 
Is there a case when `s.GCEUserFunc` is not nil here ? You just created the field and I don't see it be set anywhere else ?

```suggestion
	if s.accountEmail == "" {
		s.accountEmail = getGCEUser()
	}
```

azr: Ah, it's for tests !

azr (approved):
LGTM, thx !

PR #34, opened 2021-08-27
Treat ERROR 4047 as retryable.

xaxo05:
Treat ERROR 4047 as retryable.

There is a race condition when creating an iap tunnel to the compute instance (vm).

The code creates a vm and waits until the state of the vm is set to RUNNING. It then tries to create an iap tunnel to the vm in a loop with a timeout.

There is a period of time after the vm gets into the RUNNING state and one can actually create the tunnel. Therefore, the code tries to ignore some errors returned by the gcloud tool and retry the tunnel creation.

As of recently, we see first ERROR 4033 followed by ERROR 4047. The first one is treated as retryable, however, the second one is not. This ends all attempts to create the tunnel and thus ends the image creation with an error.

This change fixes the issue by treating ERROR 4047 returned by the gcloud tool as retryable.


azr:
Hey @yp28, this was released 3 days ago

PR #36, opened 2021-09-02
Remove dependency to Packer

azr:
Some post processors are requiring the Packer in order to get a string; to know if they can use the artifacts as a source. This creates a dependency loop between Packer and this plugin. This comments that for now. We can add this back when Packer 2.0 stops importing plugins.


sylviamoss (approved):
LGTM!! 👍🏼

PR #47, opened 2021-10-05
add source image to PARtifact

SwampDragons:
Adds source image to PARtifact. Also adds the source image to the generated Data, using the plugin sdk's generated data tool.


azr (approved):
LGTM !

PR #60, opened 2021-11-15
blog post appears to have moved

ethanmdavidson:
The previous URL doesn't seem to be in wayback machine, so I'm not 100% sure this is the right post but it seems like it's probably the same one


nywilken (approved):
Thanks for tracking down this URL. 

The domain cloudwindows.io redirects to missionimpossiblecode.io and this [blog post](https://missionimpossiblecode.io/post/rebrand-to-mission-impossible-code) seems to solidify that this is the new home for this post.

PR #65, opened 2021-12-01
update ansible example

lmayorga1980:
This will help GCP Packer Windows users to interact with Image builds using openssh


lmayorga1980:
@azr ,

Do you know if these docs make it back to packer.io site?

azr:
They will, next time we release Packer !

PR #72, opened 2021-12-13
Ensure OS Login API calls are using the correct Identities

upodroid:
Fixes: #73

Can I get some help with the `driver_mock.go`?

This change fixes a bug where the OS Login email lookup doesn't work if service account impersonation is being used. 

The `tokeninfo` function at the bottom of  `builder/googlecompute/step_import_os_login_ssh_key.go` is initialised with incorrect ClientOption and won't work if you are impersonating.

Also, `getGCEUser` is broken with service account impersonation as it uses the metadata server of the instance/pod to get the email of the identity that is impersonating and not the impersonated identity.

I have reworked it to acquire the correct Access Token from NewClientOptionGoogle and then use it to make a call to get email from https://oauth2.googleapis.com/tokeninfo

@azr @SwampDragons

### DIFF
--- a/builder/googlecompute/driver_gce.go
+++ b/builder/googlecompute/driver_gce.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	compute "google.golang.org/api/compute/v1"
+	goauth2 "google.golang.org/api/oauth2/v2"
 	"google.golang.org/api/option"
 	oslogin "google.golang.org/api/oslogin/v1"
 
@@ -33,6 +34,7 @@ type driverGCE struct {
 	projectId      string
 	service        *compute.Service
 	osLoginService *oslogin.Service
+	oauth2Service  *goauth2.Service
 	ui             packersdk.Ui
 }
 
@@ -80,25 +82,25 @@ func (ots OauthTokenSource) Token() (*oauth2.Token, error) {
 
 }
 
-func NewClientOptionGoogle(account *ServiceAccount, vaultOauth string, impersonatesa string, accessToken string) (option.ClientOption, error) {
+func NewClientOptionGoogle(account *ServiceAccount, vaultOauth string, impersonatesa string, accessToken string) ([]option.ClientOption, error) {
 	var err error
 
-	var opts option.ClientOption
+	var opts []option.ClientOption
 
 	if vaultOauth != "" {
 		// Auth with Vault Oauth
 		log.Printf("Using Vault to generate Oauth token.")
 		ts := OauthTokenSource{vaultOauth}
-		opts = option.WithTokenSource(ts)
+		opts = append(opts, option.WithTokenSource(ts))
 
 	} else if impersonatesa != "" {
-		opts = option.ImpersonateCredentials(impersonatesa)
+		opts = append(opts, option.ImpersonateCredentials(impersonatesa))
 	} else if accessToken != "" {
 		// Auth with static access token
 		log.Printf("[INFO] Using static Google Access Token")
 		token := &oauth2.Token{AccessToken: accessToken}
 		ts := oauth2.StaticTokenSource(token)
-		opts = option.WithTokenSource(ts)
+		opts = append(opts, option.WithTokenSource(ts))
 	} else if account != nil && account.jwt != nil && len(account.jwt.PrivateKey) > 0 {
 		// Auth with AccountFile if provided
 		log.Printf("[INFO] Requesting Google token via account_file...")
@@ -106,14 +108,14 @@ func NewClientOptionGoogle(account *ServiceAccount, vaultOauth string, impersona
 		log.Printf("[INFO]   -- Scopes: %s", DriverScopes)
 		log.Printf("[INFO]   -- Private Key Length: %d", len(account.jwt.PrivateKey))
 
-		opts = option.WithCredentialsJSON(account.jsonKey)
+		opts = append(opts, option.WithCredentialsJSON(account.jsonKey))
 	} else {
 		log.Printf("[INFO] Requesting Google token via GCE API Default Client Token Source...")
 		ts, err := google.DefaultTokenSource(context.TODO(), "https://www.googleapis.com/auth/cloud-platform")
 		if err != nil {
 			return nil, err
 		}
-		opts = option.WithTokenSource(ts)
+		opts = append(opts, option.WithTokenSource(ts))
 		// The DefaultClient uses the DefaultTokenSource of the google lib.
 		// The DefaultTokenSource uses the "Application Default Credentials"
 		// It looks for credentials in the following places, preferring the first location found:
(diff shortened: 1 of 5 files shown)


upodroid:
Also, does Packer implement something like AtLeastOneOf from TF? `ssh_username` should be optional if `use_os_login` is set as the code already overrides `ssh_username`.

builder/googlecompute/driver_gce.go
```diff
@@ -126,11 +128,11 @@ func NewClientOptionGoogle(account *ServiceAccount, vaultOauth string, impersona
 		// 4. On Google Compute Engine and Google App Engine Managed VMs, it fetches
 		//    credentials from the metadata server.
 		//    (In this final case any provided scopes are ignored.)
-		//
-		//    Note: (4) is not usable with OSLogin on Google Compute Engine (GCE).
-		//    The GCE service account is derived separately and used instead.
 	}
 
+	// Insert UserAgents
+	opts = append(opts, option.WithUserAgent(useragent.String(version.PluginVersion.FormattedVersion())))
```
azr, line 134: Should this be more precise, something like "packer-plugin-googlecompute-v1.2.3" ?

builder/googlecompute/driver_gce.go
```diff
@@ -126,11 +128,11 @@ func NewClientOptionGoogle(account *ServiceAccount, vaultOauth string, impersona
 		// 4. On Google Compute Engine and Google App Engine Managed VMs, it fetches
 		//    credentials from the metadata server.
 		//    (In this final case any provided scopes are ignored.)
-		//
-		//    Note: (4) is not usable with OSLogin on Google Compute Engine (GCE).
-		//    The GCE service account is derived separately and used instead.
```
azr, line 131: Could these have been a used feature ?

builder/googlecompute/driver_gce.go
upodroid: Instead of doing 
`service.UserAgent = useragent.String(version.PluginVersion.FormattedVersion())` 3 times inside `NewClientOptionGoogle()` for each service, I inserted the useragent much earlier in the auth options. Was the string incorrect before?

upodroid: Also, that UA is propagated correctly to the post-processor clients as well. 😄

azr: No I think this is correct, while reading it I just thought we could be a bit more precise.

builder/googlecompute/step_import_os_login_ssh_key_test.go
```diff
@@ -145,40 +106,6 @@ func TestStepImportOSLoginSSHKey_withGCEAndNoAccount(t *testing.T) {
...
-		t.Fatal("expected to see a public key")
-	}
-
-	sha256sum := sha256.Sum256(config.Comm.SSHPublicKey)
-	if pubKey != hex.EncodeToString(sha256sum[:]) {
-		t.Errorf("expected to see a matching public key, but got %q", pubKey)
-	}
-}
-
-func TestStepImportOSLoginSSHKey_withGCEAndAccount(t *testing.T) {
```
azr, line 163: If there is a test for it, then someone probably depends on it, do you think it could be a good idea to keep these ?

azr:
> Also, does Packer implement something like AtLeastOneOf from TF?

The internal communicator should be able to try on different types of auth methods, yes.

builder/googlecompute/driver_gce.go
upodroid: I took another a look at this and it seems that I need to rework my implementation.

gceUser was introduced in https://github.com/hashicorp/packer-plugin-googlecompute/commit/95a41be763766c46a5f404e7dc63b49570f0cf40#diff-42f2cc596979842e27b1dea3a65dc4377bd631e73fab5f51acb89231ef68f829 . That snippet was put in there to acquire the email of the service account running on Google Cloud. However, this snippet shouldn't run if service account impersonation is active.

As you can see the, access token from the metadata server is missing the email field as the VMs don't have the userinfo.email scope

```
mahamed@bastion:~$ http https://oauth2.googleapis.com/tokeninfo "Authorization: Bearer $(gcloud auth print-access-token --impersonate-service-account bootstrap@REDACTED.iam
.gserviceaccount.com)"
WARNING: This command is using service account impersonation. All API calls will be executed as [bootstrap@REDACTED.iam.gserviceaccount.com].
HTTP/1.1 200 OK
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Content-Encoding: gzip
Content-Type: application/json; charset=UTF-8
Date: Tue, 14 Dec 2021 15:23:33 GMT
Expires: Mon, 01 Jan 1990 00:00:00 GMT
Pragma: no-cache
Server: ESF
Transfer-Encoding: chunked
Vary: Origin
Vary: X-Origin
Vary: Referer
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 0
{
    "access_type": "online",
    "aud": "111995373771645511267",
    "azp": "111995373771645511267",
    "email": "bootstrap@REDACTED.iam.gserviceaccount.com",
    "email_verified": "true",
    "exp": "1639499012",
    "expires_in": "3598",
    "scope": "https://www.googleapis.com/auth/cloud-platform https://www.googleapis.com/auth/compute https://www.googleapis.com/auth/appengine.admin https://www.googleapis.com/auth/use
rinfo.email openid",
    "sub": "111995373771645511267"
}

mahamed@bastion:~$ http https://oauth2.googleapis.com/tokeninfo "Authorization: Bearer $(http http://metadata.g
oogle.internal/computeMetadata/v1/instance/service-accounts/bastion@REDACTED.iam.gserviceaccount.com/token?
scopes=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcloud-platform,https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fuserinfo.
email "Metadata-Flavor: Google" | jq .access_token -r)"
HTTP/1.1 200 OK
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Content-Encoding: gzip
Content-Type: application/json; charset=UTF-8
Date: Tue, 14 Dec 2021 15:19:44 GMT
Expires: Mon, 01 Jan 1990 00:00:00 GMT
Pragma: no-cache
Server: ESF
Transfer-Encoding: chunked
Vary: Origin
Vary: X-Origin
Vary: Referer
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 0
{
    "access_type": "online",
    "aud": "117164364499178354736",
    "azp": "117164364499178354736",
    "exp": "1639498308",
    "expires_in": "3123",
    "scope": "https://www.googleapis.com/auth/cloud-platform"
}
```

upodroid: I have reworked the logic now. gceUser function is skipped if ImpersonatedServiceAccount is set and the impersonation client option is passed in to the oauth2 service being built in this file.

upodroid:
> > Also, does Packer implement something like AtLeastOneOf from TF?
> 
> The internal communicator should be able to try on different types of auth methods, yes. Does that answer your question ?

This error should be suppressed if use_os_login is set to true.

```
Error: 1 error(s) occurred:

* An ssh_username must be specified
  Note: some builders used to default ssh_username to "root".

  on image.pkr.hcl line 53:
  (source code not available)



==> Wait completed after 10 microseconds

==> Builds finished but no artifacts were created.
```

builder/googlecompute/step_import_os_login_ssh_key_test.go
upodroid: Reinstated the test.

upodroid:
> Are you able to change that error from validation ?

That parameter isn't adjustable in this package.

https://github.com/hashicorp/packer-plugin-sdk/blob/main/communicator/config.go

azr (approved):
LGTM ! Thanks

upodroid:
Thank you

PR #89, opened 2022-03-22
fixed missing comma

jacobmammoliti:
Added the missing comma on line 281 causing the following error on `go build`:
```bash
# github.com/hashicorp/packer-plugin-googlecompute/builder/googlecompute
builder/googlecompute/driver_gce.go:281:24: syntax error: unexpected newline, expecting comma or }
```

Should have validated before initial PR but fixed now... `go build` builds successfully.


azr (approved):
Ah, good catch I remember seeing the tests pass, but looks like not !

PR #96, opened 2022-06-07
Remove unneeded Sprintf

jason-h-35:
```
jasonhatton@localhost:~/code/packer-plugin-googlecompute$ golangci-lint run
builder/googlecompute/step_start_tunnel_test.go:78:5: S1025: the argument's underlying type is a slice of bytes, should use a simple conversion instead of fmt.Sprintf (gosimple)
        if fmt.Sprintf("%s", f) != expected {
```
This PR fixes this lint.


nywilken (approved):
Looks good. Thanks for the fix.

PR #112, opened 2022-08-29
Add GHA for running acceptance test nightly

nywilken:
In an attempt to test the latest code changes against the latest released version of Packer a GitHub action is being added to run nightly acceptance tests. Authentication for this action is controlled by repository secrets, which are exposed to the Packer runner by setting the GOOGLE_APPLICATION_CREDENTIALS environment variable.

* Added a .go-version file to control the version of Go used for testing


.github/workflows/acceptance-test.yml
```diff
@@ -0,0 +1,62 @@
+#
+# This GitHub action runs acceptance tests every night.
+#
+name: "Acceptance Test"
+
+on:
+  schedule:
+    # Runs against the default branch every day at midnight
+    - cron: "0 0 * * *"
+  workflow_dispatch:
```
nywilken, line 10: This was added for testing purposes. I would like to run the action manually to ensure all is working as expected.

.github/workflows/acceptance-test.yml
```diff
@@ -0,0 +1,62 @@
...
+         credentials_json: '${{ secrets.GOOGLE_CREDENTIALS }}'
+         access_token_lifetime: '7200s'
+         export_environment_variables: false
+         create_credentials_file: true
+         cleanup_credentials: true
+      - name: Run Acceptance Tests
+        run: |
+          mkdir -p /tmp/test-results
+          make dev
+          PACKER_ACC=1 gotestsum --format=short-verbose --junitfile /tmp/test-results/gotestsum-report.xml -- -count=1 -timeout=120m -p 2 ./...
```
lbajolet-hashicorp, line 59: I may miss something, but will we do anything with the junitfile? Should it be uploaded somewhere, or do we ignore it?

lbajolet-hashicorp (commented):
I assume the manual triggering will be removed when the action is tested?
Otherwise the code looks good to me, I can approve this when the `workflow_dispatch` has been taken care of.

.github/workflows/acceptance-test.yml
nywilken: Good catch. This is copy pasta. We don't do anything with. So I would say it is safe to remove.

nywilken:
> I assume the manual triggering will be removed when the action is tested? Otherwise the code looks good to me, I can approve this when the `workflow_dispatch` has been taken care of.

Thanks all is green and working as expected.

lbajolet-hashicorp (approved):

PR #118, opened 2022-09-27
fix: replace windows-startup-script-cmd with sysprep-specialize-script-cmd

vandot:
If you use `windows-restart` second boot will fail if `windows-startup-script-cmd` has action that cannot be executed twice, like `New-NetFirewallRule`. While `sysprep-specialize-script-cmd` will be executed only at first boot.

### DIFF
--- a/docs/builders/googlecompute.mdx
+++ b/docs/builders/googlecompute.mdx
@@ -221,7 +221,7 @@ setting a valid `project_id`:
       "winrm_insecure": true,
       "winrm_use_ssl": true,
       "metadata": {
-        "windows-startup-script-cmd": "winrm quickconfig -quiet & net user /add packer_user & net localgroup administrators packer_user /add & winrm set winrm/config/service/auth @{Basic=\"true\"}"
+        "sysprep-specialize-script-cmd": "winrm quickconfig -quiet & net user /add packer_user & net localgroup administrators packer_user /add & winrm set winrm/config/service/auth @{Basic=\"true\"}"
       },
       "zone": "us-central1-a"
     }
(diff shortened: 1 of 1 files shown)


nywilken:
@vandot apologies for the delays here. I'm assigning myself to give this PR a review and understand the need for this change. By chance is there accompanying issue related to the problem you mention in the description?

vandot:
@nywilken there isn't accompanying issue related to the problem, I can create it if needed?
Proposed change is just to cover edge case if you reboot Windows machine as well.
Using twice `New-NetFirewallRule` command is not possible and it's much cleaner fix to replace `windows-startup-script-cmd` with `sysprep-specialize-script-cmd` rather then write a powershell that first checks if NetFirewallRule exists etc.
[Here](https://cloud.google.com/compute/docs/instances/startup-scripts/windows#order_of_execution_of_windows_startup_scripts) is described how and when are different Windows startup scripts executed. Checking all described cases inside docs I think it makes sense to execute scripts only once and not after each boot, thus replacing `windows-startup-script-cmd` with `sysprep-specialize-script-cmd`.

vandot:
@nywilken any updates?

nywilken:
Hi @vandot I have not had a chance to review. But the provided details is enough context to get me going. Sorry for not following up sooner.

nywilken (approved):
Apologies again for the long time to review. The changes look good to me. I left one small suggestion, which I will make for you to not delay this further. Thanks for the contribution. It is a great addition.

nywilken:
Changes merged with my additional on top. Closing PR as it has been merged.

PR #153, opened 2023-04-03
Extra block device support

### LINKED ISSUE #7: [gogolecompute] Support for more than one disk
_This issue was originally opened by @pmontanari as hashicorp/packer#8899. It was migrated here as a result of the [Packer plugin split](https://github.com/hashicorp/packer/issues/8610#issuecomment-770034737). The original body of the issue is below._

<hr>

#### Feature Description

googlecompute builder is missing the fearture to attach multiple disks (like for  for Amazon AWS: ami_block_device_mappings) even though  [Machine images ](https://cloud.google.com/compute/docs/machine-images) support it

Thanks
Patrick

### LINKED ISSUE #10: Support for google local-ssd
_This issue was originally opened by @roman-cnd as hashicorp/packer#7139. It was migrated here as a result of the [Packer plugin split](https://github.com/hashicorp/packer/issues/8610#issuecomment-770034737). The original body of the issue is below._

<hr>

Google cloud offers local ssd with superior performance https://cloud.google.com/compute/docs/disks/local-ssd, packer should support this storage option

The local ssd disks are specified when instance created.

Thank you!

lbajolet-hashicorp:
This PR adds support for specifying extra block devices when creating an instance. This is done through one or more extra `block_device` blocks in the source definition.

Closes #7 
Closes #10


builder/googlecompute/block_device.go
```diff
@@ -0,0 +1,306 @@
...
+		errs = append(errs, fmt.Errorf("Requested IOPS must be >= 10000 and <= 120000"))
+	}
+
+	if bd.IOPS != 0 {
+		switch bd.VolumeType {
+		case ZonalExtreme:
+		default:
+			errs = append(errs, fmt.Errorf("IOPS may only be specified for %q volumes", ZonalExtreme))
+		}
+	}
```
nywilken, line 219: Slight refactor to get rid of some additional conditions and group checks together. 
```go
	switch bd.VolumeType {
	case LocalScratch,
		ZonalStandard, ZonalBalanced, ZonalSSD, ZonalExtreme:
	default:
		errs = append(errs, fmt.Errorf("A valid volume type was not specified %q", bd.VolumeType))
		errs = append(errs, fmt.Errorf("%s", volumeTypeError()))
		// does it make sense to continue if we get this error? 
		return errs
	}
	
	if (bd.IOPS != 0) && (bd.VolumeType  != ZonalExtreme) {
		errs = append(errs, fmt.Errorf("IOPS may only be specified for %q volumes", ZonalExtreme))
	}
	
	if bd.IOPS != 0 && (bd.IOPS < 10000 || bd.IOPS > 120000) {
		errs = append(errs, fmt.Errorf("Requested IOPS must be >= 10000 and <= 120000"))
	}

	if bd.VolumeType == LocalScratch && bd.KeepDevice {
		errs = append(errs, fmt.Errorf("Scratch volumes cannot be kept after the instance is shutdown"))
	}

	if bd.VolumeType == LocalScratch && bd.DiskName != "" {
		errs = append(errs, fmt.Errorf("Scratch volumes cannot have a name specified."))
	}

	if bd.VolumeSize == 0 {
		errs = append(errs, fmt.Errorf("volume_size must be specified"))
	}
	
	//No need to continue checking for LocalScratch types
	if bd.VolumeType == LocalScratch {
	   return errs
	 }

	if bd.DiskName != "" && (!diskNameRegex.MatchString(bd.DiskName) || len(bd.DiskName) > 63) {
		errs = append(errs, fmt.Errorf("Disk name %q is non-compliant.", bd.DiskName))
		errs = append(errs, fmt.Errorf("The name must be 1-63 characters long and comply to the regexp '[a-z]([-a-z0-9]*[a-z0-9])?'"))
	}

       //If we've gotten here then we know VolumeType is not LocalScratch
	if bd.DiskName == "" {
		uuid, err := uuid.NewV4()
		if err != nil {
			errs = append(errs, fmt.Errorf("error creating the disk name: %s", err))
			return errs
		}
		bd.DiskName = fmt.Sprintf("packer-%s", uuid.String())
		log.Printf("[TRACE] - Set disk name as %q", bd.DiskName)
	}

```

builder/googlecompute/config.go
```diff
@@ -366,6 +375,15 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 
 	var errs *packersdk.MultiError
 
+	for i, bd := range c.ExtraBlockDevices {
+		err := bd.Prepare()
+		if err != nil {
+			errs = packersdk.MultiErrorAppend(errs, err...)
```
nywilken, line 381: If bd.Prepare fails should we skip the setting of bd.zone and c.ExtraBlockDeviecs? 
```suggestion
			errs = packersdk.MultiErrorAppend(errs, err...)
			continue
```

builder/googlecompute/driver.go
```diff
@@ -12,6 +12,9 @@ import (
 // with GCE. The Driver interface exists mostly to allow a mock implementation
 // to be used to test the steps.
 type Driver interface {
+	// CreateDisk creates a persistent disk from the specified config.
```
nywilken, line 15:
```suggestion
	// CreateDisk creates one or more persistent disks from the specified config.
```

builder/googlecompute/step_create_disks.go
```diff
@@ -0,0 +1,131 @@
...
+
+		if err != nil {
+			ui.Error(fmt.Sprintf(
+				"Error deleting disk. Please delete it manually.\n\n"+
+					"Name: %s\n"+
+					"Error: %s", gceDisk.DiskName, err))
+		} else {
+			ui.Say(fmt.Sprintf("Disk %q successfully deleted", gceDisk.DiskName))
+		}
+	}
```
nywilken, line 130: What about something like this? It looks like you are avoiding the duplication of the Ui.Error string creation but the multiple if statements and select add a bit of noise. 

```go
		if err != nil {
			ui.Error(fmt.Sprintf(
				"Error deleting disk. Please delete it manually.\n\n"+ "Name: %s\n"+ "Error: %s", gceDisk.DiskName, err))
			return
		}

		select {
		  case err = <-errCh:
		  	ui.Error(fmt.Sprintf(
				"Error deleting disk. Please delete it manually.\n\n"+ "Name: %s\n"+ "Error: %s", gceDisk.DiskName, err))
			return
		  case <-time.After(config.StateTimeout):
				err = errors.New("time out while waiting for disk to delete")
		}

		ui.Say(fmt.Sprintf("Disk %q successfully deleted", gceDisk.DiskName))
}
```

builder/googlecompute/block_device.go
```diff
@@ -0,0 +1,306 @@
...
+
+func (bd *BlockDevice) prepareDiskCreate() []error {
+	if bd.SourceVolume != "" && bd.hasDiskCreationArgs() {
+		return []error{
+			fmt.Errorf(`if source_volume is set, none of the following arguments are settable:
+* disk_name
+* volume_type
+* volume_size
+* iops
+* keep_device`),
```
nywilken, line 161:
```suggestion
			fmt.Errorf(`when specifying a source_volume the following configuration arguments can not be used:
* disk_name
* volume_type
* volume_size
* iops
* keep_device`),
```

builder/googlecompute/config.go
lbajolet-hashicorp: Good call, we definitely can

builder/googlecompute/driver.go
lbajolet-hashicorp: The driver command here only creates one disk, each `BlockDevice` represents only one disk, and this function is called for each separate disk we create.

builder/googlecompute/step_create_disks.go
lbajolet-hashicorp: I'm not sure the refactor does the trick in this case. We could however simplify `DeleteDisk` so that it only returns a channel for the error, and the other error (which happens if the request cannot be performed at all for some reason) can be forwarded through the channel.

I'll see to change that.

nywilken (approved):
LGTM

PR #154, opened 2023-04-13
docs/builder: Fix reference to SSH partial

### LINKED ISSUE #134: Docs broken include of SSH-Private-Key-File-not-required.mdx fails
Example:

image

From [Google Compute Builder](https://developer.hashicorp.com/packer/plugins/builders/googlecompute#ssh_local_tunnels)

nywilken:
Documentation is broken
image

Closes #134


lbajolet-hashicorp (approved):
Good catch!

PR #159, opened 2023-05-01
Fix docs extra disks

lbajolet-hashicorp:
When the `disk_attachment` block was introduced to the builder, the docs weren't updated, making the feature essentially invisible for users.

This commit adds a section to document those, and links for easier navigation in the docs.

### DIFF
--- a/builder/googlecompute/builder_acc_test.go
+++ b/builder/googlecompute/builder_acc_test.go
@@ -157,3 +157,24 @@ func TestAccBuilder_WithExtraPersistentDiskAndRegions(t *testing.T) {
 	}
 	acctest.TestPlugin(t, testCase)
 }
+
+func TestAccBuilder_WithMultipleDisks(t *testing.T) {
+	tmpl, err := testDataFs.ReadFile("testdata/multiple_disks.pkr.hcl")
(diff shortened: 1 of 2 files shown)


docs/builders/googlecompute.mdx
```diff
@@ -521,3 +521,20 @@ CentOS and recent Debian images have root ssh access disabled by default. Set
...
 `f1-micro` or `g1-small` to build images.
+
+## Extra disk attachments
+
+@include 'builder/googlecompute/BlockDevice.mdx'
+
+These can be defined using the [disk_attachment](#disk_attachment) block in the configuration.
+
+Note that this is an array, and therefore in HCL2 can be defined as multiple blocks, each
+one corresponding to a disk that will be attached to the instance you are booting.
```
nywilken, line 532: Would it be helpful to add an example hcl2 block here?

lbajolet-hashicorp: Done, I added one example with two disks defined in the template. Also to make sure it works well, I added an acceptance test for this case.

nywilken (approved):
Looks good. Love the tests.

PR #162, opened 2023-05-09
builder: no delete logs OSLogin ssh key if not set

### LINKED ISSUE #161: Do not emit ssh public key removal logs when using ssh private key file
#### Community Note

#### Description

I'm using this Packer plugin with OS Login and a user-specified private key file. The plugin logs correctly report that an SSH key has been imported, but they mistakenly report at the end of the build that the key has been deleted. These logs confused me and, while I believe that no key has been deleted (I can in fact still see the key through `gcloud compute os-login describe-profile`), I would like for the logs not to be emitted at all.

#### Interesting configuration

```
ssh_private_key_file = "/path/to/private/key"
ssh_username = "sa-username"
use_os_login = true
```

When this configuration is used, these logs should not be emitted:
```
2023-05-05T16:25:04Z: ==> googlecompute.cos_image: Deleting SSH public key for OSLogin...
2023-05-05T16:25:04Z:     googlecompute.cos_image: SSH public key for OSLogin has been deleted!
```

lbajolet-hashicorp:
Because of SDK updates to the `SSHKeyGen` step, we automatically set the `SSHPublicKey` in the communicator when a `SSHPrivateKeyFile` is set in the template.

This invalidates some conditions in the instance creation logic, and with the OSLogin key upload step, as both relied on the value of `SSHPublicKey` to make a decision.

To prevent this, we only upload the SSH public key when no private key file is specified in the template, thereby making the plugin behave as it had prior to updating the SDK to newer versions that v0.3.2.

Closes #161


builder/googlecompute/builder_acc_test.go
```diff
@@ -55,6 +61,76 @@ func TestAccBuilder_DefaultTokenSource(t *testing.T) {
 	acctest.TestPlugin(t, testCase)
 }
 
+// generateSSHPrivateKey generates a PEM encoded ssh private key file
+//
+// The file's deletion is the responsibility of the caller.
+func generateSSHPrivateKey() (string, error) {
```
nywilken, line 67: Nice test

I think we should also update this unit test so that public key is not set to nil. It should still pass  https://github.com/hashicorp/packer-plugin-googlecompute/blob/957770ecb4ca9f3508c0688d5421b30e7fa2f66a/builder/googlecompute/step_import_os_login_ssh_key_test.go#L200

nywilken (approved):
One small suggestion but otherwise good to go.

nywilken (approved):
One small suggestion but otherwise good to go.

builder/googlecompute/builder_acc_test.go
lbajolet-hashicorp: Good call! I've updated it with a generated private key file, and it indeed still passes

nywilken (approved):
🚀

PR #167, opened 2023-05-23
builder: leverage AutoDelete for persistent disks

### LINKED ISSUE #158: Configure auto-delete for persistent disks
#### Community Note

#### Description

The [compute SDK `AttachedDisk` struct](https://pkg.go.dev/google.golang.org/api/compute/v1#AttachedDisk) has an `AutoDelete` option for specifying whether the disk will be auto-deleted when the instance is deleted (but not when the disk is detached from the instance).

Currently only [`scratch` disk types are configured to auto-delete](https://github.com/hashicorp/packer-plugin-googlecompute/blob/main/builder/googlecompute/block_device.go#L284), but it would be nice to also configure [`persistent` disk types](https://github.com/hashicorp/packer-plugin-googlecompute/blob/main/builder/googlecompute/block_device.go#L297) to auto-delete.

#### Use Case(s)

When a compute instance created by Packer that has a persistent disk is deleted outside of the Packer cleanup process, the persistent disk is not automatically deleted and is left dangling. The automation, or user, deleting the compute instance needs to remember to also delete the persistent disk resource.

If we were able to set `AutoDelete` for persistent disks attached to compute instances via Packer, the disk would be automatically deleted by GCP when its corresponding compute instance is deleted.

#### Potential configuration

Add `AutoDelete: bd.KeepDevice,` to https://github.com/hashicorp/packer-plugin-googlecompute/blob/main/builder/googlecompute/block_device.go#L297-L305.

For example:
```go
	return &compute.AttachedDisk{
		AutoDelete:        bd.KeepDevice,
		Boot:              false,
		DeviceName:        bd.DeviceName,
		Interface:         bd.InterfaceType,
		Mode:              bd.AttachmentMode,
		DiskEncryptionKey: bd.DiskEncryptionKey.ComputeType(),
		Type:              "PERSISTENT",
		Source:            bd.SourceVolume,
	}
```

Or add a new field to [BlockDevice](https://github.com/hashicorp/packer-plugin-googlecompute/blob/main/builder/googlecompute/block_device.go#L28) to configure `AutoDelete`.

#### Potential References

Code references linked in text above.

lbajolet-hashicorp:
In previous versions of the plugin, we'd remove disks manually, one by one, after the instance was torn-down.

However, there's an AutoDelete boolean that we can set for disks at instance creation time, which takes care of deleting those for us.

This commit sets this boolean to the appropriate value so they're automatically deleted, and we keep on checking/deleting manually some of those disks after that should one fail to be deleted during this step.

This should mitigate issues with the cleanup phase for this step, and ensure disks are indeed removed.

Closes #158

### DIFF
--- a/builder/googlecompute/block_device.go
+++ b/builder/googlecompute/block_device.go
@@ -281,6 +281,19 @@ func (bd BlockDevice) generateComputeDiskPayload() (*compute.Disk, error) {
 	return payload, nil
 }
 
+// shouldAutoDelete returns whether the disk should be automatically deleted after build is done.
+func (bd BlockDevice) shouldAutoDelete() bool {
+	if bd.VolumeType == LocalScratch {
+		return true
+	}
+
+	if bd.KeepDevice {
+		return false
+	}
+
+	return true
+}
+
 func (bd BlockDevice) generateDiskAttachment() *compute.AttachedDisk {
 	if bd.VolumeType == LocalScratch {
 		return &compute.AttachedDisk{
@@ -298,6 +311,7 @@ func (bd BlockDevice) generateDiskAttachment() *compute.AttachedDisk {
 	}
 
 	return &compute.AttachedDisk{
+		AutoDelete:        bd.shouldAutoDelete(),
 		Boot:              false,
 		DeviceName:        bd.DeviceName,
 		Interface:         bd.InterfaceType,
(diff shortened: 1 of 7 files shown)


nywilken (approved):
Nicely done! LGTM

nywilken (approved):
Nicely done! LGTM

nywilken (approved):
Nicely done! LGTM

PR #168, opened 2023-05-26
cleanup github workflows

lbajolet-hashicorp:
Remove the test-example.yml workflow as it is not necessary and contravenes the new guidelines for Github Actions.


.github/workflows/acceptance-test.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         # version, because "goenv" can react to it automatically.
         run: |
           echo "Building with Go $(cat .go-version)"
-          echo "::set-output name=go-version::$(cat .go-version)"
+          echo "name=go-version::$(cat .go-version)" >>"$GITHUB_OUTPUT"
```
nywilken, line 27: It looks like you are mixing the syntax style here. 
```suggestion
          echo "go-version=$(cat .go-version)" >>"$GITHUB_OUTPUT"
```

lbajolet-hashicorp: Very true, good catch!

nywilken (approved):
LGTM

Issue #30, opened 2021-08-18
Allow pushing images to separate GCP projects


ArmaanT:
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

Right now images that are generated by modified virtual machines are created in the same project that the original VM was created in. It could be useful to be given the option to create an image in a different project. This option would allow for the creation of multiple different images environments as well as potentially restrict Packer's permissions in more secure projects.

#### Use Case(s)

A good use case for this would be any situation where you would want to limit the permissions given to Packer in a specific environment. For example, it may be desirable to give Packer permissions to only modify images in a shared project while granting it instance admin on a different project to actually create and configure virtual machines.

Another potential use case is having separate QA and production projects for images. Then Packer could be configured to push images to the correct projects.

#### Potential configuration

A new `image_project_id` option that defaults to `project_id`

PR #170, opened 2023-06-09
feat: adds image_project_id parameter for pushing images to a different GCP project

opdude:
This PR resurrects and rebases [this](https://github.com/hashicorp/packer-plugin-googlecompute/pull/31) PR which creates an image_project_id configuration option (which defaults to project_id) and allows defining a different project that an image should be created in.

Closes https://github.com/hashicorp/packer-plugin-googlecompute/issues/30

### DIFF
--- a/builder/googlecompute/step_create_image.go
+++ b/builder/googlecompute/step_create_image.go
@@ -34,7 +34,7 @@ func (s *StepCreateImage) Run(ctx context.Context, state multistep.StateBag) mul
 	if config.PackerForce && config.imageAlreadyExists {
 		ui.Say("Deleting previous image...")
 
-		errCh := driver.DeleteImage(config.ImageName)
+		errCh := driver.DeleteImage(config.ImageProjectId, config.ImageName)
 		err := <-errCh
 		if err != nil {
 			err := fmt.Errorf("Error deleting image: %s", err)
@@ -47,7 +47,7 @@ func (s *StepCreateImage) Run(ctx context.Context, state multistep.StateBag) mul
 	ui.Say("Creating image...")
 
 	imageCh, errCh := driver.CreateImage(
-		config.ImageName, config.ImageDescription, config.ImageFamily, config.Zone,
+		config.ImageProjectId, config.ImageName, config.ImageDescription, config.ImageFamily, config.Zone,
 		config.DiskName, config.ImageLabels, config.ImageLicenses, config.ImageGuestOsFeatures,
 		config.ImageEncryptionKey.ComputeType(), config.ImageStorageLocations)
 	var err error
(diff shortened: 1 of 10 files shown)


builder/googlecompute/step_create_image.go
```diff
@@ -58,7 +58,7 @@ func (s *StepCreateImage) Run(ctx context.Context, state multistep.StateBag) mul
 	}
 
 	if err != nil {
-		err := fmt.Errorf("Error waiting for image: %s", err)
+		err := fmt.Errorf("Error waiting for image: %s in", err)
```
lbajolet-hashicorp, line 61: I presume you wanted to add the project ID in this log? If this is unaccessible I suggest we keep the previous message.

lbajolet-hashicorp (approved):
Hi @opdude,

Thanks for picking this one up! The code looks good to me, I just left a singe nit on an error message, but outside from that we're good I think.

Since this requires two projects to test the feature, I don't think we can easily add an acceptance test for this feature (I'll need to check what I can do on that front), have you confirmed it worked as you expect it to?

I'm pre-approving in advance since there's not a lot to change here, will circle back after you've addressed what I commented on.

Thanks again!

builder/googlecompute/step_create_image.go
opdude: Ah this changed when debugging intention was to remove this, have reverted it to the previous :)

opdude:
> Hi @opdude,
> 
> Thanks for picking this one up! The code looks good to me, I just left a singe nit on an error message, but outside from that we're good I think.
> 
> Since this requires two projects to test the feature, I don't think we can easily add an acceptance test for this feature (I'll need to check what I can do on that front), have you confirmed it worked as you expect it to?
> 
> I'm pre-approving in advance since there's not a lot to change here, will circle back after you've addressed what I commented on.
> 
> Thanks again!

Thanks. Yeah, I've been running it successfully both with and without using the override to the destination GCP project.

lbajolet-hashicorp:
Hi @opdude,

Thanks for the quick reroll! I'll run tests and if they succeed, we can consider this done, I'll merge the PR when that happens.

Thanks again for picking this one up!

ArmaanT:
Thanks for landing this @opdude! I'm glad to see these make it into main.

bbitarelle:
Hi,

Thank you for this PR !
Just to clarify, we can now create images in another project, but not in 2 projects at the same time ?

Thanks !

opdude:
> Hi,
> 
> Thank you for this PR ! Just to clarify, we can now create images in another project, but not in 2 projects at the same time ?
> 
> Thanks !

Yes it stil only builds into a singular project but allows the building and destination project to be different.

PR #171, opened 2023-06-11
change default machine type to e2-standard-2

ledmonster:
### What
This PR will change default machine_type from `n1-standard-1` to `e2-standard-2`.

### Why

`n1-standard-1` is an old machine type and it's often not available. I confirmed with asia-northeast1-a, b, c zones.


nywilken (approved):
LGTM

PR #174, opened 2023-07-06
WIP: added support for the billing-project / quota-project parameter

bukzor-sentryio:
This change is provided on an as-is basis; I don't intend to work on it any more (since it turns out we won't use this option). As such, please feel free to close this without comment :)

At the same time, I thought I'd send this much upstream so you could decide whether anyone wants to take up the remaining work.

Anecdotally, this change does "work" -- setting the billing-project option has the intended effect of redirecting API calls to the specified project, but I have (obviously) not added any tests.

CC @MeredithAnya


nywilken:
Hi @bukzor-sentryio thanks for opening the PR. But seeing as we don't have any open issues requesting this feature we will close the PR. If we get interest on the feature we will reopen the PR and continue your work. Thanks again!

Issue #188, opened 2023-10-05
[feature]googlecompute instance template


lmayorga1980:
I would like to contribute to this repo by creating an instance template, basically is a clone of `googlecompute` but instead of saving as another gcp image, it will be saved as instance template.

https://cloud.google.com/compute/docs/instance-templates

Should i start creating another builder called `googlecomputeinstancetemplate`?

lmayorga1980:
Hi @nywilken ,

What would be your suggestion in order to support `machine-images` for google compute. Maybe a forked repo and keeping it up to date with the latest changes. The feature reuses most of the work already done by the regular google image golang code but I understand why it cannot be merged. 

Please let me know

PR #189, opened 2023-10-12
add option to create machine images

lmayorga1980:
This is just an idea to add support for machine-images in packer googlecompute plugin.

Closes https://github.com/hashicorp/packer-plugin-googlecompute/issues/188


lmayorga1980:
@nywilken , I believe `amazon-ebs` has several builders that have common features between them. I believe the route is maybe going that direction, this is an attempt to support `gcp machine images` with certain packer flags.

lmayorga1980:
@nywilken closing this PR and opened a new one with the rebased version from `main`

PR #190, opened 2023-10-12
feat: add `image_source_disk` configuration setting

ericnorris:
Hello! This pull request aims to allow a user to configure which disk is used when creating the GCE image from the packer instance. I've defaulted to the `disk_name`, which makes this backwards-compatible, but with this change you could now specify a `disk_name` from one of the `disk_attachment` blocks (#153).

We believe this functionality would be useful for a simpler way of taking an image, _not booting from it_, and adding extra data to it before producing a second image.

I'm open to changing the name of the setting or any other changes you may suggest. Thanks!

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -338,8 +338,9 @@ type Config struct {
 	// Example: "us-central1-a"
 	Zone string `mapstructure:"zone" required:"true"`
 
-	imageAlreadyExists bool
 	ctx                interpolate.Context
+	imageSourceDisk    string
+	imageAlreadyExists bool
 }
 
 func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
@@ -470,6 +471,22 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 
 	if c.DiskName == "" {
 		c.DiskName = c.InstanceName
+	}
+
+	for _, bd := range c.ExtraBlockDevices {
+		if !bd.CreateImage {
+			continue
+		}
+
+		if c.imageSourceDisk != "" {
+			errs = packersdk.MultiErrorAppend(errs, errors.New("create_image cannot be enabled on multiple disks."))
+		}
+
+		c.imageSourceDisk = bd.DiskName
+	}
+
+	if c.imageSourceDisk == "" {
+		c.imageSourceDisk = c.DiskName
 	}
 
 	if c.MachineType == "" {
--- a/builder/googlecompute/config_test.go
+++ b/builder/googlecompute/config_test.go
@@ -633,6 +633,58 @@ func TestConfigExtraBlockDevice_zone_forwarded(t *testing.T) {
 	}
 }
 
+func TestConfigExtraBlockDevice_create_image(t *testing.T) {
+	var config Config
+
+	c, _ := testConfig(t)
+
+	c["disk_attachment"] = []map[string]interface{}{
+		{
+			"volume_type":  "pd-standard",
+			"volume_size":  20,
+			"disk_name":    "second-disk",
+			"create_image": true,
+		},
+	}
+
+	_, err := config.Prepare(c)
+
+	if err != nil {
+		t.Fatalf("failed to prepare config: %#s", err)
+	}
+
+	if config.imageSourceDisk != "second-disk" {
+		t.Errorf("Expected imageSourceDisk (%q) to match second disk's disk_name (%q)", config.imageSourceDisk, "second-disk")
+	}
+}
+
+func TestConfigExtraBlockDevice_create_image_multiple(t *testing.T) {
+	var config Config
+
+	c, _ := testConfig(t)
+
+	c["disk_attachment"] = []map[string]interface{}{
+		{
+			"volume_type":  "pd-standard",
+			"volume_size":  20,
+			"disk_name":    "second-disk",
+			"create_image": true,
+		},
+		{
+			"volume_type":  "pd-standard",
+			"volume_size":  20,
+			"disk_name":    "third-disk",
+			"create_image": true,
+		},
+	}
+
+	_, err := config.Prepare(c)
+
+	if err == nil {
+		t.Fatalf("expected an error due to having multiple disks with create_image enabled, got nil")
+	}
+}
+
 // Helper stuff below
 
 func testConfig(t *testing.T) (config map[string]interface{}, tempAccountFile string) {
--- a/builder/googlecompute/step_create_image.go
+++ b/builder/googlecompute/step_create_image.go
@@ -48,7 +48,7 @@ func (s *StepCreateImage) Run(ctx context.Context, state multistep.StateBag) mul
 
 	ui.Say("Creating image...")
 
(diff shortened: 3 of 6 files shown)


builder/googlecompute/config.go
```diff
@@ -494,6 +497,10 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 		c.DiskName = c.InstanceName
 	}
 
+	if c.ImageSourceDisk == "" {
```
lbajolet-hashicorp, line 500: I think the base idea is there, but there remains some questions to be answered:

1. Can we be sure the disk we pick here exists? I think in prepare we should check that it does, and if it doesn't, we should error immediately.
2. This option makes sense for disks that are not the one we base the instance on, what if they're marked auto-delete? We should probably error if we are relying on one that gets deleted, or double-check at least that it doesn't prevent from creating the image at the end (note: looking at the code I would presume it does as the disks get created when the instance is shutdown).

lbajolet-hashicorp (changes requested):
Hi @ericnorris, 

Thanks for opening this one! I think the base framework is there, now we should add more tests to make sure the template is valid and it will work.

We should probably see to add a few tests in the `config_test.go` file, just to be sure the option gets set appropriately, and that the checks work once we added them.

Finally, after the code looks OK, I can work on adding acceptance tests. These ones run in our projects, so I'll take care of that.

Let me know if you need help with anything, thanks!

builder/googlecompute/config.go
ericnorris: I had the same idea, but I wanted to keep the PR simple to gauge a reaction before I invested more time in fleshing out the error conditions - so yeah, sounds good! I'll look into that.

And good point about auto-deleting disks, I'll look into that as well.

ericnorris:
Hey @lbajolet-hashicorp, thanks for taking a look! I've made the change you suggested to validate the `image_source_disk` setting, but I do not believe that we need to validate that the disk is not configured to auto-delete, because it seems that the disks are deleted during the "cleanup" phase, and thus post-image creation: https://github.com/hashicorp/packer-plugin-googlecompute/blob/bfa1ee26ddec60a735a320e9ffb6e38b630b8991/builder/googlecompute/step_create_disks.go#L93-L147

I'll test this again when I remember how to build and use this plugin locally :)

lbajolet-hashicorp:
Hey @ericnorris,

This cleanup routine would indeed be where we delete disks after the image is created indeed, as the cleanups are called after the build is complete, in reversed-order compared to the steps they run in.

That being said, this routine is more of a failsafe, and disks that can are also registered as `auto-delete` for the instance (refer to [this code](https://github.com/hashicorp/packer-plugin-googlecompute/blob/bfa1ee26ddec60a735a320e9ffb6e38b630b8991/builder/googlecompute/block_device.go#L314) for more details), which are automatically deleted by GCE when the image is torn-down, which happens before we create the image, so in this case the disk will be deleted before the image is created.
I would suggest that we amend the code that sets the `auto-delete` flag to not include it if the disk we pick is the base of the image we're creating.

ericnorris:
Ah, I see what you mean. At a glance this is kinda tricky though, because `shouldAutoDelete` has no knowledge of the rest of the config as far as I can tell. I'll look into plumbing that information through, but feel free to point me in a specific direction if you have an idea!

lbajolet-hashicorp:
Yes, since `shouldAutoDelete` is only using the disk's config to determine if it should be deleted, it cannot know if it's the target for the image or not.
I would maybe suggest we add a flag to the extra disks (private so it doesn't end-up as part of the expected config), and set it for the disk that is to become the image? I'm open to other ideas if you have them!

ericnorris:
@lbajolet-hashicorp yep, I was just thinking that! Alternatively, we could move the setting to the `disk_attachment` block proper and actually make it a part of the config. For example:

```hcl
disk_attachment {
  # ...
  create_image true
}
```

Here we could either (a) create multiple images, one for each time `create_image` is true, or (b) only allow this to be specified once. If we went with (a), I'd probably want to also add a setting for skipping image creation for the boot disk.

I'm fine with any of the ideas. (a) would likely be the most work, but also the most flexible; the other two (your suggestion or (b)) are equally easy I think.

ericnorris:
@lbajolet-hashicorp for the sake of expediency, I've gone ahead and implemented the private flag option, so let me know if this works for you. Otherwise, I'm happy to implement any of the other options. Thanks!

builder/googlecompute/config.go
```diff
@@ -178,6 +178,9 @@ type Config struct {
 	//  }
 	//  ```
 	ImageStorageLocations []string `mapstructure:"image_storage_locations" required:"false"`
+	// The name of the disk that should be used to create an image. Defaults to the
+	// value of the `disk_name` setting if not set.
+	ImageSourceDisk string `mapstructure:"image_source_disk" required:"false"`
```
lbajolet-hashicorp, line 183: Come to think of it, this only makes sense if one of the extra disks is to become the image source, and it's supposedly one only.
Should this be a flag on one of the extra disks? Something like `is_image_source` maybe?

Basically this amounts to what you've added as a private flag to the structure, but instead making it public and settable in the configs.

Then we'd need to check if we have this flag set for one disk, and propagate the created disk as the target for the image creation step.

What do you think? Let me know if the approach sounds good to you, and if you need a hand getting this in, I can help, just say the word :)

lbajolet-hashicorp (commented):
Hi @ericnorris,

Sorry for taking long in between the reroll and this review.

I took another look at the code, and I think this works, but I wonder if it would not be simpler to tag one of the extra disks as the target through a flag than it is to repeat the disk name in the global configs.

Please let me know what you think, and wave if you need me to pick-up the work here, happy to help.

Thanks again for the contribution, much appreciated!

ericnorris:
That makes sense to me, and I believe what you're suggesting is what I was trying to express in https://github.com/hashicorp/packer-plugin-googlecompute/pull/190#issuecomment-1780058320 - we could expose `create_image` as a flag on the disk, and either a) image every disk that has this flag, which is the most flexible option, or b) allow only one disk to have that and to use that as the source for the image.

What you're proposing sounds like option (b), so I'll take a stab at implementing that and see how it looks!

lbajolet-hashicorp:
> That makes sense to me, and I believe what you're suggesting is what I was trying to express in [#190 (comment)](https://github.com/hashicorp/packer-plugin-googlecompute/pull/190#issuecomment-1780058320) - we could expose `create_image` as a flag on the disk, and either a) image every disk that has this flag, which is the most flexible option, or b) allow only one disk to have that and to use that as the source for the image.
> 
> What you're proposing sounds like option (b), so I'll take a stab at implementing that and see how it looks!

Oh! Good call, I completely missed that comment, apologies!
Thanks for the quick turnaround here, and thanks again for being willing to contribute!

ericnorris:
No worries @lbajolet-hashicorp! I started from scratch due to the merge conflicts, and so I've force-pushed the new approach. I can try to actually test this locally at some point in the near future.

lbajolet-hashicorp (approved):
Hi @ericnorris,

Thanks for the reroll, the code in this state looks great to me, I'll merge this ASAP.

Thanks again for pushing through with this feature!

PR #197, opened 2023-11-09
Common code reorganisation

lbajolet-hashicorp:
This PR reworks the code for all the components, now instead of having the common code as part of the builder, and have the post-processors depend on it, we moved all the common code to a `lib` package, which is then leveraged by components.

This removes some duplication in the authentication logic, and consolidates it in a single package, along with the driver, which gets more capabilities for the post-processors to use.

This however means (and this was already the case in the related PR) that the `googlecompute-export` component now needs to expose its own authentication methods, as it cannot rely on the artifact for that.

Leaving authentication information in the artifact may be a security issue later, so we opted to remove that.

Note: based on top of #186


JenGoldstrich (approved):
LGTM! Nice refactor

lmayorga1980:
This will be very nice to support `machine images`

lbajolet-hashicorp:
Not sure what the linter is complaining about, locally this doesn't fail so I'll override the test and merge this as-is, we'll fix the issue if it becomes persistent.

PR #198, opened 2023-11-10
Fix acceptance tests with local SSDs

lbajolet-hashicorp:
Some machine types don't support local SSDs as extra block devices, so we change our acceptance tests that contain those so they use n2-* machines instead of E2-*, which don't support those block device attachments.


nywilken (approved):
Great catch!

PR #207, opened 2024-02-02
Docs fixes

lbajolet-hashicorp:
While taking a look at the plugin's documentation, I noticed that some structures were not referenced anywhere, making it hard to figure out the full capabilities of the plugin without looking at examples or the code directly.

Also, when PR #186 was merged, some components had migrated to the `lib/common` package, but the old doc partials were not removed (packer-sdc does not remove existing partials), and their includes were still present in the docs.

This PR fixes both issues.


nywilken (approved):
Good catch!

PR #211, opened 2024-03-01
add omit_external_ip and use_internal_ip fields to export post-processor

cmgsj:
After trying to use the `googlecompute-export` post-processor, it fails when provisioning the exporter instance because it violates the organization policy `constraints/compute.vmExternalIpAccess`.

This is the `packer` code used:

```hcl
source "googlecompute" "vm" {
  // ...
}

build {
  name = "main"
  sources = ["sources.googlecompute.vm"]
  // ...
  post-processor "googlecompute-export" {
    paths = [
      "gs://gs://$BUCKET_NAME/$IMAGE_NAME.tar.gz",
    ]
    keep_input_artifact = true
  }
}
```

and a fragment of the build logs:

```text
==> main.googlecompute.vm: Running post-processor: (type googlecompute-export)
==> main.googlecompute.vm (googlecompute-export): Exporting image $IMAGE_NAME to destination: [gs://$BUCKET_NAME/$IMAGE_NAME.tar.gz]
==> main.googlecompute.vm (googlecompute-export): Creating temporary RSA SSH key for instance...
==> main.googlecompute.vm (googlecompute-export): Using image: debian-9-worker-v20200616
==> main.googlecompute.vm (googlecompute-export): Creating instance...
    main.googlecompute.vm (googlecompute-export): Loading zone: $ZONE
    main.googlecompute.vm (googlecompute-export): Loading machine type: n1-highcpu-4
    main.googlecompute.vm (googlecompute-export): Requesting instance creation...
==> main.googlecompute.vm (googlecompute-export): Error creating instance: googleapi: Error 412: Constraint constraints/compute.vmExternalIpAccess violated for project $PROJECT_NUMBER. Add instance projects/$PROJECT_ID/zones/$ZONE/instances/$IMAGE_NAME-exporter to the constraint to use external IP with it., conditionNotMet
```

This happens because the provisioned export instance always has the fields `omit_external_ip` and `use_internal_ip` set to `false`, with no way to overwrite them.

This PR **attempts** to change that behavior, by exposing the aforementioned fields in the `googlecompute-export` post-processor config and then using them in the `googlecompute.Config` of the export instance.

Example:

```hcl
source "googlecompute" "vm" {
  // ...
}

build {
  name = "main"
  sources = ["sources.googlecompute.vm"]
  // ...
  post-processor "googlecompute-export" {
    paths = [
      "gs://gs://$BUCKET_NAME/$IMAGE_NAME.tar.gz",
    ]
    keep_input_artifact = true
    omit_external_ip    = true
    use_internal_ip     = true
  }
}
```

Please let me know if a way to circumvent this issue already exists. Any other feedback is welcome as well 🙂!

### DIFF
--- a/post-processor/googlecompute-export/post-processor.go
+++ b/post-processor/googlecompute-export/post-processor.go
@@ -68,6 +68,12 @@ type Config struct {
 	Zone                string `mapstructure:"zone"`
 	IAP                 bool   `mapstructure-to-hcl2:",skip"`
 	ServiceAccountEmail string `mapstructure:"service_account_email"`
+	//If true, the export instance will not have an external IP. use_internal_ip must
+	//be true if this property is true.
+	OmitExternalIP bool `mapstructure:"omit_external_ip" required:"false"`
+	//If true, use the export instance's internal IP instead of its external IP
+	//during exporting.
+	UseInternalIP bool `mapstructure:"use_internal_ip" required:"false"`
 
 	ctx interpolate.Context
 }
@@ -183,6 +189,8 @@ func (p *PostProcessor) PostProcess(ctx context.Context, ui packersdk.Ui, artifa
 		SourceImageProjectId: []string{"compute-image-tools"},
 		Subnetwork:           p.config.Subnetwork,
 		Zone:                 p.config.Zone,
+		OmitExternalIP:       p.config.OmitExternalIP,
+		UseInternalIP:        p.config.UseInternalIP,
 		Scopes: []string{
 			"https://www.googleapis.com/auth/compute",
 			"https://www.googleapis.com/auth/devstorage.full_control",
--- a/.web-docs/components/post-processor/googlecompute-export/README.md
+++ b/.web-docs/components/post-processor/googlecompute-export/README.md
@@ -74,6 +74,12 @@ the [authentication](/packer/integrations/hashicorp/googlecompute#authentication
 
 - `service_account_email` (string) - Service Account Email
 
+- `omit_external_ip` (bool) - If true, the export instance will not have an external IP. use_internal_ip must
+  be true if this property is true.
+
+- `use_internal_ip` (bool) - If true, use the export instance's internal IP instead of its external IP
+  during exporting.
+
 <!-- End of code generated from the comments of the Config struct in post-processor/googlecompute-export/post-processor.go; -->
 
 
--- a/docs-partials/post-processor/googlecompute-export/Config-not-required.mdx
+++ b/docs-partials/post-processor/googlecompute-export/Config-not-required.mdx
@@ -35,4 +35,10 @@
 
 - `service_account_email` (string) - Service Account Email
 
+- `omit_external_ip` (bool) - If true, the export instance will not have an external IP. use_internal_ip must
+  be true if this property is true.
+
+- `use_internal_ip` (bool) - If true, use the export instance's internal IP instead of its external IP
+  during exporting.
+
 <!-- End of code generated from the comments of the Config struct in post-processor/googlecompute-export/post-processor.go; -->
--- a/post-processor/googlecompute-export/post-processor.hcl2spec.go
+++ b/post-processor/googlecompute-export/post-processor.hcl2spec.go
@@ -33,6 +33,8 @@ type FlatConfig struct {
 	Subnetwork                *string           `mapstructure:"subnetwork" cty:"subnetwork" hcl:"subnetwork"`
 	Zone                      *string           `mapstructure:"zone" cty:"zone" hcl:"zone"`
 	ServiceAccountEmail       *string           `mapstructure:"service_account_email" cty:"service_account_email" hcl:"service_account_email"`
+	OmitExternalIP            *bool             `mapstructure:"omit_external_ip" required:"false" cty:"omit_external_ip" hcl:"omit_external_ip"`
+	UseInternalIP             *bool             `mapstructure:"use_internal_ip" required:"false" cty:"use_internal_ip" hcl:"use_internal_ip"`
 }
 
 // FlatMapstructure returns a new FlatConfig.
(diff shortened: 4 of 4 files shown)


post-processor/googlecompute-export/post-processor.go
```diff
@@ -68,6 +68,12 @@ type Config struct {
 	Zone                string `mapstructure:"zone"`
 	IAP                 bool   `mapstructure-to-hcl2:",skip"`
 	ServiceAccountEmail string `mapstructure:"service_account_email"`
+	//If true, the export instance will not have an external IP. use_internal_ip must
+	//be true if this property is true.
+	OmitExternalIP bool `mapstructure:"omit_external_ip" required:"false"`
+	//If true, use the export instance's internal IP instead of its external IP
+	//during exporting.
+	UseInternalIP bool `mapstructure:"use_internal_ip" required:"false"`
```
nywilken, line 76: Looking at the usage of this attribute it would appear that we only need one attribute `use_internal_ip` as you can not have one be true without the other. Is that correct? 

```suggestion
	//By default the external IP used by the exporting machine have an external accessible IP.
	//If you need to suppress the use of an external IP for security reasons set this configuration to `true`
	// to force the export instance its internal IP instead of its external IP during exporting.
	UseInternalIP bool `mapstructure:"use_internal_ip" required:"false"`
```

post-processor/googlecompute-export/post-processor.go
```diff
@@ -183,6 +189,8 @@ func (p *PostProcessor) PostProcess(ctx context.Context, ui packersdk.Ui, artifa
 		SourceImageProjectId: []string{"compute-image-tools"},
 		Subnetwork:           p.config.Subnetwork,
 		Zone:                 p.config.Zone,
+		OmitExternalIP:       p.config.OmitExternalIP,
+		UseInternalIP:        p.config.UseInternalIP,
```
nywilken, line 193: My understanding is as follows
1. When UseInternalIP is true OmitExternalIP must be True. 
1. When OmitExternalIP is true UseInternalIP must be True. 
1. When UseInternalIP is false OmitExternalIP must be false. 
1.  OmitExternalIP can not be false when UseInternal is True
1. UseInternalIP can not be false when OmitExternalIP is true because of case 2. 

```suggestion
		OmitExternalIP:       p.config.UseInternalIP,
		UseInternalIP:        p.config.UseInternalIP,
```

nywilken (changes requested):
Hi there apologies for the delayed review here. Our focus has been on Packer itself trying to resolve issues with plugin loading. Looking at the changes and reference issue I can understand the need for this feature. 

I left a few suggestions on the change based on my understanding on how Omit and Use internal are working. Please let me know if you have any questions or if my understanding is off.

cmgsj:
@nywilken thank you so much for taking the time to review!

I initially thought of adding both fields `use_internal_ip` and `omit_external_ip` in order to be consistent with the fields the `googlecompute` source exposes.

I have identified three possible successful combinations of the above fields according to the following truth table:

| case | **use_internal_ip** | **omit_external_ip** | **result**                                                    |
| ---- | ------------------- | -------------------- | ------------------------------------------------------------- |
| 1    | `true`              | `true`               | ✅ internal IP is used, instance does not have an external IP |
| 2    | `true`              | `false`              | ✅ internal IP is used, instance has an external IP           |
| 3    | `false`             | `true`               | ❌ cannot use external IP because instance does not have one  |
| 4    | `false`             | `false`              | ✅ external IP is used, instance has an external IP           |

All the statements you made [above](https://github.com/hashicorp/packer-plugin-googlecompute/pull/211#discussion_r1530967704) are `true` except the following two, which are indeed **possible**, but not **necessary**:
- ***When UseInternalIP is true OmitExternalIP must be True.***
- ***OmitExternalIP can not be false when UseInternal is True.***

I think you might not have taken case `2` into account. 

Having said that, I do think that your suggestion is very reasonable and will prevent any potential "build-time" errors from case `3`, although it would also prevent case `2`.

In the end I think the disjunction is between allowing all cases (with a the potential error case `3`), or allowing only cases `1` and `4` (without the potential success case `2`).

Please let me know if you agree, and/or what your suggestion would be, thanks in advance!

cmgsj:
@nywilken whenever you have some time, please let me know if my [comment](https://github.com/hashicorp/packer-plugin-googlecompute/pull/211#issuecomment-2008300399) above makes more sense, or if you think it would be better to go forward with the changes you had originally suggested.

Thanks.

cmgsj:
@nywilken I pushed a [commit](https://github.com/hashicorp/packer-plugin-googlecompute/pull/211/commits/4fa2ddf5ff396e29c87d75a81a9691cdbf84010d) for printing an error on case `#3` in my [comment](https://github.com/hashicorp/packer-plugin-googlecompute/pull/211#issuecomment-2008300399) above.

Please review it when you have the time. Thanks.

nywilken:
@cmgsj Im no longer working on these PRs. Pinging @lbajolet-hashicorp to continue pushing this forward. Please keep in mind that his response may be slow as they work on different priorities.

cmgsj:
Hi @lbajolet-hashicorp,

I've submitted this PR with some enhancements to the `googlecompute-export` post-processor, and I'd appreciate it if you could take a look whenever you have a moment.

Let me know if you have any questions or need additional context. Thanks in advance for your time!

lbajolet-hashicorp (changes requested):
Hi @cmgsj,

Sorry for the delay in the review here, I was busy working on other things, we're planning a release for the GCE plugin this week, so I'm going through the opened PRs now to see what we can include in it.

Regarding your change, while I understand the reason to add `OmitExternalIP` to the `googlecompute-export` post-processor here, as it's used by the step that creates the temporary instance, the `UseInternalIP` seems superfluous here.

`UseInternalIP` is only used by `StepInstanceInfo`, which while used in the builder for GCE for populating the instance's IP address to use for the remainder of the build steps. It is not used in the export post-processor at all since there's not a lot done in there, so I would think you can remove it, remove the extra check for both attributes set, and amend the comment so that `OmitExternalIP` is the only attribute added here.

Of course I might be missing something so please let me know if that's the case.

Marking as request changes for now, but hopefully this can be quickly resolved, thanks!

lbajolet-hashicorp:
Oh, on another note, I see the CLA has not been accepted by your Github account yet, which means we cannot accept the contribution until this is done. Could you do this at the same time so we aren't blocked if/when it comes to merging this PR?

cmgsj:
@lbajolet-hashicorp thanks for the review!

The changes you suggested make much more sense, I have implemented them.
I realize that `UseInternalIP` is only used by the builder's `StepInstanceInfo`.

I also fixed the CLA issue.

lbajolet-hashicorp (approved):
Hey again @cmgsj,

Looks much simpler indeed :)

With that it seems good to me, we can merge it right away, thanks for the quick reroll, much appreciated!

PR #214, opened 2024-03-22
Image architecture config

BrennenMM7:
This PR is to add image_architecture as a config for the packer plugin. Currently this plugin only supports provisioning Image Types of X86_64 in GCP disabling the ability to create ARM based images. 

image_architecture will allow consumers of this plugin to modify the disk type being provisioned by GCE to enable ARM based packer images in GCP. 

Test and validation are included with this PR to ensure that image_architecture is set to X86_64 or ARM64, by default it will select X86_64 as its the most common. This field is also optional so leaving it blank or an empty value will serve backwards compatibility for current configs.

### DIFF
--- a/builder/googlecompute/config.go
+++ b/builder/googlecompute/config.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"regexp"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/packer-plugin-googlecompute/lib/common"
@@ -115,6 +116,8 @@ type Config struct {
 	IAPConfig `mapstructure:",squash"`
 	// Skip creating the image. Useful for setting to `true` during a build test stage. Defaults to `false`.
 	SkipCreateImage bool `mapstructure:"skip_create_image" required:"false"`
+	// The architecture of the resulting image. Defaults to "x86_64".
+	ImageArchitecture string `mapstructure:"image_architecture" required:"false"`
 	// The unique name of the resulting image. Defaults to
 	// `packer-{{timestamp}}`.
 	ImageName string `mapstructure:"image_name" required:"false"`
@@ -443,16 +446,28 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 		}
 	}
 
+	// Set the default image architecture to X86_64.
+	if c.ImageArchitecture == "" {
+		c.ImageArchitecture = "X86_64"
+	}
+
+	// Ensure the image architecture is uppercase
+	c.ImageArchitecture = strings.ToUpper(c.ImageArchitecture)
+
 	// used for ImageName and ImageFamily
 	imageErrorText := "Invalid image %s %q: The first character must be a lowercase letter, and all following characters must be a dash, lowercase letter, or digit, except the last character, which cannot be a dash"
+	if !validImageName.MatchString(c.ImageName) {
+		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf(imageErrorText, "name", c.ImageName))
+	}
 
 	if len(c.ImageName) > 63 {
 		errs = packersdk.MultiErrorAppend(errs,
 			errors.New("Invalid image name: Must not be longer than 63 characters"))
 	}
 
-	if !validImageName.MatchString(c.ImageName) {
-		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf(imageErrorText, "name", c.ImageName))
+	if !(c.ImageArchitecture == "X86_64" || c.ImageArchitecture == "ARM64") {
+		errs = packersdk.MultiErrorAppend(errs,
+			errors.New("Invalid image architecture: Must be either X86_64 or ARM64"))
 	}
 
 	if len(c.ImageFamily) > 63 {
--- a/builder/googlecompute/config_test.go
+++ b/builder/googlecompute/config_test.go
@@ -64,7 +64,6 @@ func TestConfigPrepare(t *testing.T) {
 			"foo",
 			false,
 		},
-
 		{
 			"zone",
 			nil,
@@ -682,6 +681,35 @@ func TestConfigExtraBlockDevice_create_image_multiple(t *testing.T) {
 
 	if err == nil {
 		t.Fatalf("expected an error due to having multiple disks with create_image enabled, got nil")
+	}
+}
+
+func TestConfigPrepareImageArchitecture(t *testing.T) {
+	cases := []struct {
+		Architecture string
+		ExpectValid  bool
+	}{
+		{"X86_64", true},
+		{"ARM64", true},
+		{"arm64", true},
+		{"x86_64", true},
+		{"i386", false}, // Assuming 'i386' is not supported, adjust as needed
+	}
+
+	for _, tc := range cases {
+		raw, tempfile := testConfig(t)
+		defer os.Remove(tempfile)
+
+		raw["image_architecture"] = tc.Architecture
+
+		var c Config
+		_, errs := c.Prepare(raw)
+
+		if tc.ExpectValid && errs != nil {
+			t.Errorf("Expected architecture '%s' to be valid, but got error: %s", tc.Architecture, errs)
+		} else if !tc.ExpectValid && errs == nil {
+			t.Errorf("Expected architecture '%s' to be invalid, but got no error", tc.Architecture)
+		}
 	}
 }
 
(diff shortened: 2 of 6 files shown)


builder/googlecompute/config_test.go
```diff
@@ -685,17 +684,45 @@ func TestConfigExtraBlockDevice_create_image_multiple(t *testing.T) {
...
-		"credentials_file": tempAccountFile,
-		"project_id":       "hashicorp",
-		"source_image":     "foo",
-		"ssh_username":     "root",
-		"image_family":     "bar",
+		"credentials_file":   tempAccountFile,
+		"project_id":         "hashicorp",
+		"source_image":       "foo",
+		"ssh_username":       "root",
+		"image_architecture": "",
```
lbajolet-hashicorp, line 724: I'd think you don't need to change that, string are always the empty string by default

builder/googlecompute/config_test.go
```diff
@@ -685,17 +684,45 @@ func TestConfigExtraBlockDevice_create_image_multiple(t *testing.T) {
 	}
 }
 
+func TestConfigPrepareImageArchitecture(t *testing.T) {
+	cases := []struct {
+		Architecture string
+		Valid        bool
```
lbajolet-hashicorp, line 690: Super micro nit: I would name this `ExpectValid` just to be clear that this is the test's expectation and not a parameter.

builder/googlecompute/config.go
```diff
@@ -443,16 +445,24 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 		}
 	}
 
+	if c.ImageArchitecture == "" {
+		c.ImageArchitecture = "x86_64"
```
lbajolet-hashicorp, line 449: I wonder if this should default to x86_64, we need to double check regarding the values for `Image.Insert` in the API docs (doesn't seem to be indicated at first glance), but I wonder if we shouldn't leave it empty be default.
My rationale is that if the origin image's architecture setting is forwarded (from your description it doesn't seem to be, but I'd like to confirm), we'll break templates that may exist already.

lbajolet-hashicorp (commented):
Hi @BrennenMM7,

Thanks for the PR, this will definitely be useful when released.

I left a couple of nits/suggestions on the code, please feel free to address/discuss those, and we can iterate on that.

I think this is the kind of feature that would benefit from acceptance tests, I'll see what I can do once we've stabilised the base code.
Obviously if you are willing to implement some acceptance test for this feature this would be very appreciated, and don't hesitate to reach out if you need a hand with those, I'll be happy to help.

lbajolet-hashicorp:
Also, since you change the documentation for the `Config` structure, its docs partials need to be regenerated. You can use the `generate` target from the Makefile to do so if you have `make` installed on your machine.

builder/googlecompute/config.go
BrennenMM7: I think by default when this setting is nil or set to a blank string the GCP API defaults it to X86_64, only if this setting is explicitly set to a different string value would it change the templating? This is at least what I've seen in my local testing of building images.

builder/googlecompute/config_test.go
BrennenMM7: This has been renamed to the suggested value.

builder/googlecompute/config_test.go
BrennenMM7: I removed the var as its not required.

builder/googlecompute/config.go
```diff
@@ -443,16 +445,24 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
...
+	}
 
 	if len(c.ImageName) > 63 {
 		errs = packersdk.MultiErrorAppend(errs,
 			errors.New("Invalid image name: Must not be longer than 63 characters"))
 	}
 
-	if !validImageName.MatchString(c.ImageName) {
-		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf(imageErrorText, "name", c.ImageName))
+	if !(c.ImageArchitecture == "x86_64" || c.ImageArchitecture == "ARM64") {
```
lbajolet-hashicorp, line 463: I wonder, the APIs are capitalising x86_64 as `X86_64`, does that need to be exactly this, or is it case-insensitive?
If everything's case-sensitive and upper-case, may I suggest we process non-empty image architectures to be uppercase (`strings.ToUpper` should do the trick)
Please feel free to disregard if the API is case-insensitive though!

builder/googlecompute/config.go
```diff
@@ -115,6 +115,8 @@ type Config struct {
 	IAPConfig `mapstructure:",squash"`
 	// Skip creating the image. Useful for setting to `true` during a build test stage. Defaults to `false`.
 	SkipCreateImage bool `mapstructure:"skip_create_image" required:"false"`
+	// The architecture of the resulting image. Defaults to "x86_64".
+	ImageArchitecture string `mapstructure:"image_architecture" required:"false"`
```
lbajolet-hashicorp, line 119: I'm starting to think about this, but since we're running the builder as an instance on GCE, could we maybe infer the image architecture from the instance type if it is unset, rather than defaulting to `x86_64`? This may not fit all cases to be fair though, so there might be an argument to be made in favour of defaulting to x86_64 since that's what we've supported until now.

Edit: been looking at the API docs, we could maybe lift stuff from the disks of the instance at first glance (the instance itself doesn't expose its architecture), but this happens later in the process so I'm not sure this is a clean way to infer that from. The `machineTypes` endpoint does not return the architecture of a machine as well (bit weird if you ask me). I presume we can default to x86_64 for now, this looks like the saner default

lbajolet-hashicorp (commented):
Hi @BrennenMM7,

Thanks for the reroll! I left a couple extra comments, especially regarding the case for the architecture, I'm not sure the APIs are case-sensitive for that enum, but it might be a good idea to make sure of that before merging (the acceptance tests may help to figure this one out though, I'll see if I can run them).

Other than that, this LGTM!

builder/googlecompute/config.go
BrennenMM7: I've added an else to when evaluating the default value, if the user has specified a specific image arch, the string will be auto converted to uppercase in the event it was not properly set.

builder/googlecompute/config_test.go
```diff
@@ -685,6 +684,33 @@ func TestConfigExtraBlockDevice_create_image_multiple(t *testing.T) {
 	}
 }
 
+func TestConfigPrepareImageArchitecture(t *testing.T) {
+	cases := []struct {
+		Architecture string
+		ExpectValid  bool
+	}{
+		{"x86_64", true},
+		{"ARM64", true},
```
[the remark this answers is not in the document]
BrennenMM7: I modified the test case value to be upper case, however did you still want me to have uppercase testing here? I'm unsure if this would be required as only two values are valid and both are uppercase in string.

builder/googlecompute/config.go
```diff
@@ -443,16 +446,27 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 		}
 	}
 
+	// Set the default image architecture to X86_64. If the user has specified a value ensure it is uppercase.
+	if c.ImageArchitecture == "" {
+		c.ImageArchitecture = "X86_64"
+	} else {
```
lbajolet-hashicorp, line 452: I believe we don't even need the `else` statement here, regardless what the value was it can't hurt to capitalise it even if it's not necessary.

builder/googlecompute/config_test.go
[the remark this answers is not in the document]
lbajolet-hashicorp: The code is simple enough that we may not need it here, but I tend to err in favour of making sure we don't break it later, so if you're in for a final reroll to address the `else` in the `Prepare` function, I'd suggest adding one case with like `x86_64` or `arm64` to be sure they're not rejected, but if you don't no stress, that's not a blocker for merging afterwards I'd think

builder/googlecompute/config.go
BrennenMM7: I've removed the else, to have it uppercase irregardless of input.

BrennenMM7:
@lbajolet-hashicorp Any chance of a green thumb? 🟢

lbajolet-hashicorp (approved):
Hi @BrennenMM7,

Thanks for the reminder, and sorry I completely forgot to check back in after you re-uploaded the code, apologies for that!

I did a quick last review, the code looks good to me, I believe we can merge this today, thanks for being patient with us!

PR #218, opened 2024-04-21
Fix miss makezero bug

alingse:
## Change

When I was adding features for linter [makezero](https://github.com/ashanbrown/makezero), I ran some GitHub Actions for some repos on GitHub, and it reported an error in this project.

## Tests

it was a bug, I don't know how to add a test for it


lbajolet-hashicorp (approved):
Hi @alingse,

Good catch from `makezero`, it does seem that these slices have empty data given the `append` below. It's a bit of a wonder this never was reported before, though I guess the GCE APIs remove empty metadata instead of erroring?

Anyway thanks for the fix, will merge!

PR #221, opened 2024-06-04
all: remove usage of ioutil

lbajolet-hashicorp:
Since the ioutil package is deprecated, we update all the instances in which it is used to replace them with the preferred methods of executing those functions.

### DIFF
--- a/builder/googlecompute/config_test.go
+++ b/builder/googlecompute/config_test.go
@@ -5,7 +5,6 @@ package googlecompute
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"runtime"
 	"strings"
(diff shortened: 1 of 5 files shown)


builder/googlecompute/builder_acc_test.go
```diff
@@ -284,3 +284,31 @@ func TestAccBuilder_WithMultipleDisks(t *testing.T) {
 	}
 	acctest.TestPlugin(t, testCase)
 }
+
+// Note: will be disabled unless an extra envvar is set to avoid timeouting
```
JenGoldstrich, line 288: I think this test could be added in a different PR, maybe also in a different file, since its not something we'd run in most instances, why do you want to add it in this PR?

lbajolet-hashicorp: Oh yeah no that's a mistake on my part here, it should not have been part of this PR.
I'll remove it now.

JenGoldstrich (approved):

PR #226, opened 2024-07-11
Fix zone or region detection regex

gac3k:
Currently, there are some zones (a very few but still) in GCP that are not conforming to that regular expression due to having two digits in the region name, before the zone suffix, i.e. europe-west10.

That PR is to allow both variations to be supported, otherwise a disk is not properly removed when an instance is being removed, and has to be deleted manually, as the script treats the value as region name not the zone.

### DIFF
--- a/lib/common/block_device.go
+++ b/lib/common/block_device.go
@@ -245,7 +245,7 @@ func GetRegionFromZone(zone string) (string, error) {
 	return matches[1], nil
 }
 
-var zoneRegexp = regexp.MustCompile("^[a-z]+-[a-z]+[0-9]-[a-z]$")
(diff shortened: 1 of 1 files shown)


lbajolet-hashicorp (commented):
Hi @gac3k,

Thanks for the PR, LGTM! Looks like there might be some CLA issue with the commits here, have you committed them with your Github account email?
Any chance you can double-check this? We unfortunately cannot accept the PR if the CLA's not signed for legal reasons.

gac3k:
@lbajolet-hashicorp I think there was a slight issue originally, but then I resign it, and now it looks fine the checks section (?)

lbajolet-hashicorp (approved):
Hey @gac3k,

Thanks for the update, it does seem it's gone green so I'll merge this now!

PR #227, opened 2024-07-16
builder: fix support credentials files

silentsokolov:
Fix the step "start tunnel" after refactoring [849dee](https://github.com/hashicorp/packer-plugin-googlecompute/commit/849deeecfa3ea309b03fcd3f8f87a0187d328995)
Because the warning "_account_file is deprecated, please use either credentials_json or credentials_file instead_" was confusing and caused misunderstanding.


lbajolet-hashicorp (approved):
Hey @silentsokolov, good catch!

Thanks for the PR, merging now!

PR #228, opened 2024-07-22
lib: add hyperdisk to supported volume types

lbajolet-hashicorp:
Hyperdisks are another volume type with high-performance compared to other persistent options.
Since they weren't supported by the builder, we add those volume types to the configuration so users can provision their volumes with those.


nywilken (approved):
LGTM

PR #192, opened 2023-10-23
builder: add fedora-coreos-cloud project

lbajolet-hashicorp:
When a user specifies a `image_family_name` to get the latest version of an image to build their custom image upon, the plugin attempts to get the family specified from a static list of projects.

In this case, a user attempted to get the `fedora-coreos-stable' family, present in the `fedora-coreos-cloud` project, which was not in the default list of the GCE driver.

This commit changes this.

Closes: #191

PR #230, opened 2024-07-24
builder: add deeplearning-platform-release project (issue #229)

### LINKED ISSUE #229: Cannot use source images from deeplearning-platform-release project
#### Overview of the Issue

Similar to https://github.com/hashicorp/packer-plugin-googlecompute/issues/191, source images from the `deeplearning-platform-release` project cannot be used with this plugin.
Related: https://github.com/hashicorp/packer-plugin-googlecompute/issues/193

#### Reproduction Steps

Use a source image from the `deeplearning-platform-release` project, such as `common-cu123-v20240708-debian-11`

### Plugin and Packer version

Packer: v1.11.0
packer-plugin-googlecompute: v1.1.3_x5.0_linux_amd64

```
{
  ...
  "builders": [
    {
      "type": "googlecompute",
      "source_image": "common-cu123-v20240708-debian-11"
      ...
    }
}
```

Linux x86-64

```
Error getting source image for instance creation: Could not find image, common-cu123-v20240708-debian-11, in projects,
[<our project>, almalinux-cloud centos-cloud cloud-hpc-image-public cos-cloud coreos-cloud debian-cloud fedora-cloud fedora-coreos-cloud freebsd-org-cloud-dev rhel-cloud rhel-sap-cloud rocky-linux-cloud suse-cloud suse-sap-cloud suse-byos-cloud ubuntu-os-cloud windows-cloud windows-sql-cloud gce-nvme google-containers opensuse-cloud ubuntu-os-pro-cloud ml-images]
```

jbaldassari:
Similar to the issue/changes described in https://github.com/hashicorp/packer-plugin-googlecompute/pull/192, this adds support for pulling source images from the `deeplearning-platform-release` [project](https://cloud.google.com/deep-learning-vm/docs/images#listing-versions).

Closes #229

### DIFF
--- a/lib/common/driver_gce.go
+++ b/lib/common/driver_gce.go
@@ -435,6 +435,7 @@ func (d *driverGCE) GetImage(name string, fromFamily bool) (*Image, error) {
 		"opensuse-cloud",
 		"ubuntu-os-pro-cloud",
 		"ml-images",
+		"deeplearning-platform-release",
 	}
 	return d.GetImageFromProjects(projects, name, fromFamily)
 }


lbajolet-hashicorp:
Hi @jbaldassari,

Thanks for the PR! I wonder though, I missed that option last time I changed/merged code into the driver regarding the default projects, and if this is a Google official project, I can see us adding it to the list (haven't been able to confirm that unfortunately, would appreciate if you could share a link to the GCE docs if that's the case?), that being said, there's the `source_image_project_id` argument you can use to specify extra projects to look images from, does that allow you to get the image in the current state?

Just trying to limit the amount of projects we add to our default list, as it is quite rigid to update, and the more projects we add to that list, the slower (potentially) searching for a specific image/image family may become.

That being said the code looks good to me, if there's a compelling reason to add this project to the default list, we should be able to merge this PR, and we have a release planned for next week, so we can definitely fold this in then.

jbaldassari:
Thanks for the info, @lbajolet-hashicorp !  I'm sorry we missed that `source_image_project_id` option when we were looking for workarounds.  I've confirmed that setting `source_image_project_id=deeplearning-platform-release` does work for us in this case.

> if this is a Google official project, I can see us adding it to the list (haven't been able to confirm that unfortunately, would appreciate if you could share a link to the GCE docs if that's the case?),

I think the confusion here is that these images are part of Deep Learning, which is a Google Cloud product, but these images are not part of GCE itself.  The Deep Learning docs reference the `deeplearning-platform-release` project [here](https://cloud.google.com/deep-learning-vm/docs/images#listing-versions) and [here](https://cloud.google.com/deep-learning-containers/docs/choosing-container).  That being said, since we have a workaround, I'll leave it up to you if you want to include this as one of the default projects or recommend people use `source_image_project_id` for less common cases like this.

lbajolet-hashicorp:
Don't be sorry @jbaldassari, I also missed it last time when merging the PR you referenced, so there's probably something we can do to make it more visible :sweat_smile:

I'll keep the PR open for now, and bring that up next time we discuss open issues with the rest of the team, we'll make a decision then.

Thanks for the update, glad `source_image_project_id` did the trick!

lbajolet-hashicorp:
Hi @jbaldassari,

Just leaving a comment here to let you know that as discussed previously, since there's a workaround available for this problem, I'll close the PR now, along with the related issue.

Please feel free to reopen if there's anything I missed, thanks again!

Issue #232, opened 2024-07-30
Breaking change with `image_architecture` in 1.1.5


rifelpet:
#### Overview of the Issue

The recent 1.1.5 release included a breaking change in https://github.com/hashicorp/packer-plugin-googlecompute/pull/214.

When unset the value defaults to `X86_64`. In prior versions, the value implicitly defaulted to the machine_type's architecture. This means when building with an `ARM64` machine_type and source_image, upgrading to 1.1.5 introduces this error:

`Error waiting for image: googleapi: Error 400: Invalid value for field 'resource.architecture': 'X86_64'. Requested architecture must be the same as the source resource architecture (ARM64)., invalid`

The error is fixed by setting `image_architecture = "ARM64"`. This required change is not mentioned in the release notes. 

I think a better default for the `image_architecture` field is to use the machine_type's architecture, since it would be one less field that users need to add when building ARM64 images.

#### Reproduction Steps

Build an image from an ARM64 machine type and source image but without image_architecture set.

### Plugin and Packer version

`Packer v1.11.2`

### Simplified Packer Buildfile

The below configuration fails to build. Adjusting the `required_plugins` to `1.1.4` succeeds.

```hcl
packer {
  required_plugins {
    googlecompute = {
      source  = "github.com/hashicorp/googlecompute"
      version = "= 1.1.5"
    }
  }
}

source "googlecompute" "example" {
  source_image_family = "ubuntu-minimal-2204-lts-arm64"
  machine_type        = "t2a-standard-1"
  ssh_username        = "ubuntu"
  project_id          = "..."
  image_name          = "..."
  zone                = "..."
  subnetwork          = "..."
}

build {
  name    = "ansible"
  sources = ["source.googlecompute.example"]
}
```

### Operating system and Environment details

OS, Architecture, and any other information you can provide about the
environment.

### Log Fragments and crash.log files

Here are logs from a successful build with 1.1.4:

```
ansible.googlecompute.example: output will be in this color.

==> ansible.googlecompute.example: Checking image does not exist...
==> ansible.googlecompute.example: Creating temporary RSA SSH key for instance...
==> ansible.googlecompute.example: no persistent disk to create
==> ansible.googlecompute.example: Using image: ubuntu-minimal-2204-jammy-arm64-v20240725
==> ansible.googlecompute.example: Creating instance...
    ansible.googlecompute.example: Loading zone: us-central1-a
[...]

PR #234, opened 2024-08-07
Image arch

lbajolet-hashicorp:
This PR fixes issue #232 where the default value for `image_architecture` (introduced in v1.1.5 of the plugin) defaulted to x86_64, thereby breaking some existing arm64 build configs.
With this PR, we default again to leaving it empty, so the source image architecture gets forwarded to the final image being produced.
We also add some acceptance tests for this, so in the future we have an architecture to test those templates, and make sure we don't break it accidentally again.

Closes: #232


builder/googlecompute/builder_acc_test.go
```diff
@@ -284,3 +286,101 @@ func TestAccBuilder_WithMultipleDisks(t *testing.T) {
 	}
 	acctest.TestPlugin(t, testCase)
 }
+
+func TestAccBuilder_ImageArchVariations(t *testing.T) {
```
JenGoldstrich, line 290: As I understand this this adds four new executions of builds to GCP, I am not sure of the existing status of parallelizing these tests but it may be worth investigating when adding this many tests, also it looks like acceptance tests are currently failing https://github.com/hashicorp/packer-plugin-googlecompute/actions/runs/10275873966/job/28435328892

builder/googlecompute/config.go
```diff
@@ -116,7 +116,8 @@ type Config struct {
 	IAPConfig `mapstructure:",squash"`
 	// Skip creating the image. Useful for setting to `true` during a build test stage. Defaults to `false`.
 	SkipCreateImage bool `mapstructure:"skip_create_image" required:"false"`
-	// The architecture of the resulting image. Defaults to "x86_64".
+	// The architecture of the resulting image. Defaults to the empty string,
+	// the original image's architecture will carry over by default.
```
JenGoldstrich, line 120:
```suggestion
	// GCP will use original image's architecture will carry over by default.
```

I feel like this makes it more clear that we're not sending the default architecture, Google just uses it when unset

JenGoldstrich (approved):
LGTM largely, left some minor feedback, I think it'd be good to get the acceptance tests passing before releasing here though

builder/googlecompute/builder_acc_test.go
lbajolet-hashicorp: Addressed in a subsequent PR

PR #235, opened 2024-08-07
Acctest fixes and parallel

lbajolet-hashicorp:
This PR does a couple of things:

1. Fix acceptance tests: a bunch of acctests were failing because of an old image being used, so we bump them so the can succeed.
2. Run acceptance tests on push to main: instead of a cron, we now run when pushing a commit to main. This will reduce the amount of total runs, and enable us to receive alerts if they failed.
3. Parallelise acceptance tests: right now we don't have a lot of them, but we still have a non-negligible number which run sequentially, each taking 60-90s. This adds up, and leads to acceptance tests being quite long, so we enabled parallelisation for running them, which locally brought down the time to run all of them to ~2m30s.


builder/googlecompute/builder_acc_test.go
```diff
@@ -328,18 +348,21 @@ func TestAccBuilder_ImageArchVariations(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			imageName := fmt.Sprintf("%s-%d", tt.name, time.Now().UTC().Unix())
+		testRun := tt
```
JenGoldstrich, line 351: Why do you create this separate object? Can't you just use tt everywhere?

JenGoldstrich (approved):
LGTM! That was fast, nice one Lucas!

builder/googlecompute/builder_acc_test.go
lbajolet-hashicorp: Ah, very good question! Because the test runs in parallel, the captured object will be replaced by the next item, so I need to capture it locally with this in order for the value to be what it was when the coroutine is created.
I would've been caught by this tbh, `go vet` saved me there by pointing it out

PR #240, opened 2024-12-04
builder: Adds a check for image labels validity for googlecompute builder

DocEmmetBrown:
The image labels are only checked by GCP once the image is built, and will fail the build if they are not valid.
This adds a check so that the build can fail early.

Labels requirement docs is [here](https://cloud.google.com/compute/docs/labeling-resources)

Thanks for your review.


builder/googlecompute/step_create_image.go
```diff
@@ -46,6 +46,13 @@ func (s *StepCreateImage) Run(ctx context.Context, state multistep.StateBag) mul
 		}
 	}
 
+	if !config.AreLabelsValid() {
+		err := fmt.Errorf("Image: %s has invalid labels. The keys and values must start with a lowercase character, can only contain lowercase letters, numeric characters, underscores and dashes. They can be at most 63 characters long. International characters are allowed.", config.ImageName)
```
lbajolet-hashicorp, line 50: This is good, but I wonder if we can be more directive regarding the kind of error message we return our users there. Error messages like those do point out the problem, but I think we can pinpoint which key/value is non-compliant there instead of returning this generic error.

If `configs.AreLabelsValid` returns an `[]error`, we can populate that collection with the exact key/value that is non-compliant, leading to easier troubleshooting.
What do you think?

lbajolet-hashicorp (commented):
Hi @DocEmmetBrown,

Thanks for the PR, and sorry I didn't take a look at it before today.

Overall this looks good, but I think we can improve on the error reporting so that we list the labels that are non-compliant, so users have an easier time fixing them.

Thanks again!

builder/googlecompute/step_create_image.go
DocEmmetBrown: @lbajolet-hashicorp: Agreed that it would be a much better User Experience. I'll give it a go when time permits, unless you want to do it :)

lbajolet-hashicorp: As you prefer, my goal was to release a new version of the plugin this week, so the earlier we can get this in, the better :)
If you think you'll have time to address that change, then I'll let you do this, but otherwise feel free to let me know, and I'll take a jab at that!

DocEmmetBrown: Just gave it a go. Happy to get feedback

lbajolet-hashicorp: Awesome, many thanks @DocEmmetBrown ! I'll review this ASAP

lbajolet-hashicorp (approved):
Hi again @DocEmmetBrown,

The reroll looks good to me, thanks for pushing it through, much appreciated!

Assuming tests are green, I'll be merging this now!

chrisstaite:
The label regex is incorrect. It states that labels must start with with a lowercase letter, but this is incorrect. That restriction only applies to keys.

https://cloud.google.com/compute/docs/labeling-resources#requirements

PR #242, opened 2024-12-11
Add instance_max_run_duration_seconds config 

phoenixuprising:
This pr adds a new `instance_max_run_duration_seconds` config field
that gets piped down into the builder instance's `maxRunDuration`
schedule. This will allow for an instance to automatically be stopped
or deleted, even if the packer build process is killed.

It also adds an `instance_termination_action` which must be set if
`instance_max_run_duration_seconds` config is set. This will tell GCP
to either `STOP` or `DELETE` the instance once the max duration has
been reached.


phoenixuprising:
@nywilken - I saw you commenting on another PR. Is this something you can look at or pass off to someone that can?

My team is running packer in a github action and people will cancel the action sometimes while it is running, which doesn't give packer the opportunity to clean up instances correctly.

nywilken:
> @nywilken - I saw you commenting on another PR. Is this something you can look at or pass off to someone that can?
> 
> My team is running packer in a github action and people will cancel the action sometimes while it is running, which doesn't give packer the opportunity to clean up instances correctly.

@phoenixuprising I’m no longer working on this project. But I’m cc’ing @lbajolet-hashicorp to help get eyes on the pull-request. Please be mindful that the team’s priority might be elsewhere at the moment. So he made need time to review and respond.

builder/googlecompute/config.go
```diff
@@ -632,6 +637,18 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
 		c.WindowsPasswordTimeout = 3 * time.Minute
 	}
 
+	if c.InstanceMaxRunDurationSeconds < 0 {
```
nywilken, line 640: My suggestion here would be to include a high default value and possibly validate the the provided value is not lower than the default timeouts set for SSH or other Packer processes. To avoid terminating the machine before establishing a connection. It might not be so straightforward to know all the timeouts so I would suggest documenting it as well. Within the configuration option that at minimum the value should be set to X number of seconds. Possibly even hours or days.


I’ll leave this question for @lbajolet-hashicorp Should this be an unsigned int?

builder/googlecompute/config.go
```diff
@@ -632,6 +637,18 @@ func (c *Config) Prepare(raws ...interface{}) ([]string, error) {
...
+		errs = packersdk.MultiErrorAppend(fmt.Errorf("'instance_max_run_duration_seconds' must be positive"))
+	}
+
+	if c.InstanceMaxRunDurationSeconds > 0 && c.InstanceTerminationAction == "" {
+		errs = packersdk.MultiErrorAppend(fmt.Errorf("'instance_termination_action' must be set if 'instance_max_run_duration_seconds' is set"))
+	}
+
+	if c.InstanceTerminationAction != "" && !(c.InstanceTerminationAction == "STOP" || c.InstanceTerminationAction == "DELETE") {
```
nywilken, line 648: It might be worth adding constants for these values.

that aside would it make sense to  defaulting to STOP after x days.

nywilken (commented):
This is a great idea and is a nice feature add. I left a few comments but Im no longer on the team so Packeristas may have other comments.

builder/googlecompute/config.go
phoenixuprising: It seems that if it is not straightforward to know all the timeouts, setting a high default without checking the other config values would be the best path forward to me. I could set the default to something like 2 days and add documentation to the config values. I also like the idea of setting the default termination option to `STOP` so I'll add that.

> Should this be an unsigned int?
Probably. I don't believe that `maxRunDuration` can be negative.

phoenixuprising:
@lbajolet-hashicorp - Just wanted to check in and see if you could look over this PR. Thanks.

builder/googlecompute/config_test.go
```diff
@@ -712,6 +712,72 @@ func TestConfigPrepareImageArchitecture(t *testing.T) {
 	}
 }
 
+func TestConfigInstanceMaxRunDurationSeconds(t *testing.T) {
+	cases := []struct {
+		Keys   []string
+		Values []interface{}
+		Err    bool
```
lbajolet-hashicorp, line 719: Very nitpicky but in terms of self-documentation, it would imho be better if this field was called `ExpectErr` instead of just `Err` to convey the intent that this is the expected outcome of the test here.

builder/googlecompute/config.go
lbajolet-hashicorp: Yeah regarding the timeout situation I tend to agree with @phoenixuprising, it will be tricky to infer a good value here, especially since things can be set from other attributes, and having something that is coherent and long-term viable is troublesome.
I do agree that this could be a `uint`, but I assume the rationale for it being an `int64` is because it is the type expected by the SDK/API? I don't mind this being either personally, but if we're dealing with a signed number, I would suggest adding a sanity check during `Prepare` to make sure it is always positive, as is the case in the current state of the code.
Note: an argument could still be made in favour of a `uint` here, as then the type explicitly prevents negative values, so later changes in the codebase bear less risk of unintentional breakage, ultimately the choice is yours @phoenixuprising.

I like the idea of having a high enough default value, you suggest 2 days, I would even go lower and suggest something like 1 day, I would think this is reasonable enough for the vast majority of our users, and will fix that problem automagically for everyone.
Regarding the default termination option, I wonder if `STOP` is the right default here though, my read on this is that we want to reduce cost if a build gets stuck for too long, therefore I would maybe go with `DELETE` as the default option, so the instance is wiped if this happens, as to not incur extra cost.

lbajolet-hashicorp: Note: if we pick a non-zero default value here, we should probably amend the documentation for the attribute so the default value is clearly stated, along with a way to disable that behaviour.

lbajolet-hashicorp (commented):
Hi @phoenixuprising,

Thanks for the PR and apologies for the delay in reviewing.

The feature looks good, I do think it will be broadly useful for our users, good call!

I left a few comments that largely echo what was previously discussed, please let me know how you plan to address them, and we can move forward from here.

Overall this LGTM besides those though, looking forward to hearing from you on those comments!

PR #249, opened 2025-02-13
builder: fix image label sanity checks

### LINKED ISSUE #248: Version `1.1.7` blocks valid image label values
#### Overview of the Issue

Since version `1.1.7`, likely due to #240, we are unable to publish images that have a label **value** that starts with a number:

```
Image: <redacted> has invalid labels: value "380ab351648ae3591f65fa9fb7521df0c432000e" for key "git-
sha" must match regex "^[a-z][a-z0-9_-]{0,62}$",
```

I believe this is overly restrictive - the documentation (https://cloud.google.com/compute/docs/labeling-resources) states (emphasis mine):

> Keys and values can contain only lowercase letters, numeric characters, underscores, and dashes. All characters must use UTF-8 encoding, and international characters are allowed. **Keys** must start with a lowercase letter or international character.

Keys must start with a lowercase letter, but not values. In our case, the value is rejected for not starting with a letter, which I think is incorrect. We are unable to publish images using this plugin version due to this.

#### Reproduction Steps

1. Build an image with a label key of `git-sha`, and a label value of `380ab351648ae3591f65fa9fb7521df0c432000e`

### Plugin and Packer version

`plugin-googlecompute_v1.1.7_x5.0_linux_amd64`

N/A

N/A

N/A

lbajolet-hashicorp:
When the image sanity checks were introduced in v1.1.7, an error slipped in as the same regex was used to validate keys and labels, however the rules are slightly different between the two, as keys are mandated to start with [a-z], while values are not.

This commit addresses this bug by changing to using two regexes, so we can still validate that both are valid before attempting to build the image, but we are not unnecessarily rejecting valid values in the map.

Closes #248


chrisstaite (changes requested):
The new regex allows empty label values which is invalid.

lbajolet-hashicorp:
@chrisstaite that is surprising, according to the GCP [docs](https://cloud.google.com/compute/docs/labeling-resources#requirements), it does state that `Values can be empty, and have a maximum length of 63 characters.`, is this not accurate?

chrisstaite (commented):
You also assume lowercase ASCII (i.e. a-z) whereas the requirements from Google state *lowercase letters* implying lowercase non-latin characters are also allowed, specifically in UTF-8 encoding.

lbajolet-hashicorp:
You do make a point regarding lowercase letters though, I wonder what is the exact character set they allow there, I assume we can use a Unicode-friendly lowercase collation algorithm for those, but a regexp supporting those is likely tricky.
I can think of another way to do this kind of check, but in the meantime I wonder if reverting the original commit may not be better. It has the undesirable effect of rejecting an image after it's been built, but it's better than preventing builds until then.

I'll amend the PR to do that.

chrisstaite:
You should be able to use the regex `\p{Ll}` although I've not tested this

lbajolet-hashicorp:
I might add that even the length check is somewhat broken in this case, as the length of the string according to their documentation is up-to 63 characters, not 63 bytes, so the length checks are likely broken too here.

lbajolet-hashicorp:
You are right though, it seems that `\p{Ll}` might work here, assuming the repetition works on the characters and not bytes (which I would assume is the case), we should be good to use that.
I'll experiment with that a bit, and if I cannot make it work easily, I'll resort to reverting the commit instead.

lbajolet-hashicorp:
@chrisstaite thanks for the pointer to `\p{Ll}`, this seems to be working with international lower-case characters now, hopefully this won't be a problem since I worry that between unicode versions there might be incompabtibilities in terms of determining what is a lower-case letter, though those blocks are fairly stable these days and only the astral plan tends to be updated, so I assume this will be good enough in the current state.
If you still have time, might I ask you to test it once more, and give it a quick second review?

jkoelndorfer:
I encountered #248 today, also. It appears that this validation happens after the image build process.

If this validation is going to be performed by Packer, it would be much more useful for it to happen _before_ the image build starts. If the build process has already completed, at that point you might as well just ask GCP to create the image. If there is an error, GCP's API will say so and that can be reported to the user.

lbajolet-hashicorp:
@jkoelndorfer this PR performs this check during the `Prepare` step, which happens before the builder starts, so this is indeed run before anything else. This was the intent with the change that was rolled into v1.1.7 as well, I mistakenly assumed that the check was done during this step, but it was done during the `step_create_image`, which as you point out, does not prevent the plugin from doing the actual build beforehand, that was my mistake, apologies for this.

If you test this change on your build and an invalid key/value, you will see that the error is immediate now.

chrisstaite (approved):

builder/googlecompute/config.go
```diff
@@ -657,24 +663,17 @@ func ApplyIAPTunnel(c *communicator.Config, port int) error {
 	}
 }
 
+var labelKeyRegex = regexp.MustCompile(`^\p{Ll}(\p{Ll}|[0-9_-]){0,62}$`)
```
DocEmmetBrown, line 666: If I'm not mistaken, you should be able to slightly simplify the regex: 
```suggestion
var labelKeyRegex = regexp.MustCompile(`^\p{Ll}[\p{Ll}0-9_-]{0,62}$`)
```

builder/googlecompute/config.go
```diff
@@ -657,24 +663,17 @@ func ApplyIAPTunnel(c *communicator.Config, port int) error {
...
+var labelValueRegex = regexp.MustCompile(`^(\p{Ll}|[0-9_-]){0,63}$`)
```
DocEmmetBrown, line 667:
```suggestion
var labelValueRegex = regexp.MustCompile(`^[\p{Ll}0-9_-]{0,63}$`)
```

DocEmmetBrown (approved):
Thanks a lot @lbajolet-hashicorp for quickly fixing the bug I introduced 🙏

builder/googlecompute/config.go
lbajolet-hashicorp: Yeah that could work if the grammar allows us using character classes in an interval, but that's easily testable, I'll try that locally, and if it works, we can roll with it

lbajolet-hashicorp: OK I tested locally, seems to be working alright indeed, thanks for the suggestion, we should be good to go with this change!

mogrogan (approved):
LGTM !

PR #265, opened 2025-04-18
Update latest guestOsFeatures for GCP

abhay2101:
**DELETE THIS TEMPLATE BEFORE SUBMITTING**

In order to have a good experience with our community, we recommend that you
read the contributing guidelines for making a PR, and understand the lifecycle
of a Packer Plugin PR:

https://github.com/hashicorp/packer-plugin-googlecompute/blob/main/.github/CONTRIBUTING.md#opening-an-pull-request

Describe the change you are making here!
Add additional guesos features added recently in gcp images
https://cloud.google.com/compute/docs/images/create-custom#guest-os-features

Please include tests. We recommend looking at existing tests as an example. 
Build image with  TDX_CAPABLE and it shows in describe image for gcp.

If your PR resolves any open issue(s), please indicate them like this so they will be closed when your PR is merged:

Closes #xxx
Closes #xxx

### DIFF
--- a/.web-docs/components/post-processor/googlecompute-import/README.md
+++ b/.web-docs/components/post-processor/googlecompute-import/README.md
@@ -70,7 +70,7 @@ the [authentication](/packer/integrations/hashicorp/googlecompute#authentication
 
 - `image_guest_os_features` ([]string) - A list of features to enable on the guest operating system. Applicable only for bootable images. Valid
   values are `MULTI_IP_SUBNET`, `UEFI_COMPATIBLE`,
-  `VIRTIO_SCSI_MULTIQUEUE`, `GVNIC` and `WINDOWS` currently.
+  `VIRTIO_SCSI_MULTIQUEUE`, `GVNIC`, `WINDOWS`, `IDPF`, `SEV_CAPABLE`, `SEV_SNP_CAPABLE`, `SEV_LIVE_MIGRATABLE_V2`,  `TDX_CAPABLE` and `SUSPEND_RESUME_COMPATIBLE` currently.
 
 - `image_labels` (map[string]string) - Key/value pair labels to apply to the created image.
 
--- a/docs-partials/post-processor/googlecompute-import/Config-not-required.mdx
+++ b/docs-partials/post-processor/googlecompute-import/Config-not-required.mdx
@@ -23,7 +23,7 @@
 
 - `image_guest_os_features` ([]string) - A list of features to enable on the guest operating system. Applicable only for bootable images. Valid
   values are `MULTI_IP_SUBNET`, `UEFI_COMPATIBLE`,
-  `VIRTIO_SCSI_MULTIQUEUE`, `GVNIC` and `WINDOWS` currently.
+  `VIRTIO_SCSI_MULTIQUEUE`, `GVNIC`, `WINDOWS`, `IDPF`, `SEV_CAPABLE`, `SEV_SNP_CAPABLE`, `SEV_LIVE_MIGRATABLE_V2, `TDX_CAPABLE` and `SUSPEND_RESUME_COMPATIBLE` currently.
 
 - `image_labels` (map[string]string) - Key/value pair labels to apply to the created image.
 
--- a/post-processor/googlecompute-import/post-processor.go
+++ b/post-processor/googlecompute-import/post-processor.go
@@ -60,7 +60,7 @@ type Config struct {
 	ImageFamily string `mapstructure:"image_family"`
 	//A list of features to enable on the guest operating system. Applicable only for bootable images. Valid
 	//values are `MULTI_IP_SUBNET`, `UEFI_COMPATIBLE`,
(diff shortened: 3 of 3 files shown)


tanmay-hc:
@abhay2101 Thanks for the PR! Adding the `TDX_CAPABLE` flag to the image is a good first step, but on its own it won’t actually launch a TDX-enabled VM. We also need to enable confidential compute at instance launch time for it to work as expected. Could you update the PR to include that part too?

abhay2101:
> @abhay2101 Thanks for the PR! Adding the `TDX_CAPABLE` flag to the image is a good first step, but on its own it won’t actually launch a TDX-enabled VM. We also need to enable confidential compute at instance launch time for it to work as expected. Could you update the PR to include that part too?

Yes we need to add "--confidential-compute-type=TDX "below during launch time and also add TDX_CAPABLE  in "image_guest_os_features" in packer file.
 
You want me to  update instance create readme at https://github.com/search?q=repo%3Ahashicorp%2Fpacker-plugin-googlecompute%20%22instances%20create%22&type=code ? Reason didn't updated that coz not everyone want to create confidential VM's.

tanmay-hc:
@abhay2101 The PR seems to just be updating the docs and adding information about support for TDX in the ImageGuestOSFeatures. I highly doubt this alone would do the job. We might have to update the code to enable the confidential computing as well.

If this alone is working for you, would you be able to update the PR with some testing details?

abhay2101:
> @abhay2101 The PR seems to just be updating the docs and adding information about support for TDX in the ImageGuestOSFeatures. I highly doubt this alone would do the job. We might have to update the code to enable the confidential computing as well.
> 
> If this alone is working for you, would you be able to update the PR with some testing details?

@tanmay-hc Updated PR with testing details. thanks.

tanmay-hc:
@abhay2101 I tried using the test you mentioned and as expected, using just `TDX_CAPABLE` in the `image_guest_os_features` doesn't make the instance TDX enabled. As mentioned by you as well, we would need to support the `confidential-compute-type=CONFIDENTIAL_COMPUTING_TECHNOLOGY` param in Packer.

abhay2101:
> @abhay2101 I tried using the test you mentioned and as expected, using just `TDX_CAPABLE` in the `image_guest_os_features` doesn't make the instance TDX enabled. As mentioned by you as well, we would need to support the `confidential-compute-type=CONFIDENTIAL_COMPUTING_TECHNOLOGY` param in Packer.

Just to make it more clearer:
1.  we need to have ""image_guest_os_features": "UEFI_COMPATIBLE,VIRTIO_SCSI_MULTIQUEUE,GVNIC,TDX_CAPABLE,SEV_CAPABLE,SEV_SNP_CAPABLE","  in the packer json/HCL to make image to support TDX or other confidential feature.
2. Now when instance boots up in gcp that time we need to have "confidential-compute-type=CONFIDENTIAL_COMPUTING_TECHNOLOGY`" in gcloudcli(console have bug and doesn't support this) to have instance boot with TDX. this is different step and nothing to do with packer build.
3. I have tested this on RHEL9.5

tanmay-hc:
@abhay2101 our approach is valid and should work fine in most cases. That said, I think it makes sense for us to add a new builder flag — `confidential_compute = true` — which would set:
```
ConfidentialInstanceConfig: &compute.ConfidentialInstanceConfig{
	EnableConfidentialCompute: true,
}
```

This would let users opt into building the image inside a Confidential VM if needed, which can help with OS setups that behave differently in that environment.

abhay2101:
> @abhay2101 our approach is valid and should work fine in most cases. That said, I think it makes sense for us to add a new builder flag — `confidential_compute = true` — which would set:
> 
> ```
> ConfidentialInstanceConfig: &compute.ConfidentialInstanceConfig{
> 	EnableConfidentialCompute: true,
> }
> ```
> 
> This would let users opt into building the image inside a Confidential VM if needed, which can help with OS setups that behave differently in that environment.
> 
> Also, please do a `make generate` to update the new documentation changes.

@tanmay-hc  Updated documentation via "make generate". thanks.

tanmay-hc:
@abhay2101 Would you also be able to make the `EnableConfidentialCompute` changes? What are your thoughts on this?

abhay2101:
> @abhay2101 Would you also be able to make the `EnableConfidentialCompute` changes? What are your thoughts on this?

@tanmay-hc  Not fully convinced that we need this change coz 1.  tdx will only get enabled on certain hw and not supported by all. 2. building TDX enabled image doesn't need tdx enabled OS/HW. It is only needed when we boot tdx enabled guest OS. 3. works fine now with current changes.

we might not gain much here with adding that flag.

tanmay-hc:
@abhay2101 I agree the flag addresses a niche use case. I'm merging this PR for now and will consider adding the flag later if a concrete use case arises.

tanmay-hc (approved):

PR #268, opened 2025-04-23
Updated AccTest Github Action to upload failure logs as artifacts

tanmay-hc:
Adding a job to upload failure logs as artefacts to debug better in case an acceptance test fails


anurag5sh (approved):
LGTM!

PR #272, opened 2025-05-12
Adds configurable OS Login SSH username support

### LINKED ISSUE #254: OSLogin doesn't correctly determine organization external username
I'm trying to create a GCE VM image in my GCP test organization using the example below:
```
source "googlecompute" "basic-example2" {
  project_id            = "my-project-id"
  source_image          = "rhel-8-v20250123"
  machine_type          = "n2-standard-4"
  zone                       = "us-central1-a"
  image_name            = "my-image-rhel8-v20250303"
  image_family          = "my0image-rhel8"
  image_description     = "Customized RHEL 8 VM Image"
  use_os_login          = true
  ssh_username          = "ext_my_user_my_org_domain"
  network               = "my-network"
  subnetwork            = "my-network-us-central1"
  use_internal_ip       = true
  omit_external_ip      = true
  service_account_email = "my-vm@my-project-id.iam.gserviceaccount.com"
  use_iap               = true
}

build {
  sources = ["sources.googlecompute.basic-example"]
}
```

No matter what I try I am getting 

> Build 'googlecompute.basic-example' errored after 3 minutes 11 seconds: Packer experienced an authentication error when trying to connect via SSH. This can happen if your username/password are wrong. You may want to double-check your credentials as part of your debugging process. original error: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain

This is unexpected because I can SSH into the remote instance while Packer is retrying from the same machine (my local laptop) in the adjacent bash shell session using "gcloud compute ssh ..." without problems.

I suspect that the problem is that my GCP user that I'm logged in locally (and for application default crednetials as well) is coming NOT from the GCP organization where my target project is and where Packer remote VM is running.

Hence, when I'm logging in into that VM, my user name is  `**ext_**my_user_my_org_domain` (please note the standard `ext_` prefix). While when Packer is running in `-debug` mode, it shows that it ignores the `ssh_username` field I'm setting and picks the wrong username `my_user_my_org_domain` (no `ext_` prefix) when connecting to the remote VM:

> ==> googlecompute.basic-example: Obtaining SSH Username for OSLogin...
>     googlecompute.basic-example: ssh_username: my_user_my_org_domain
> 

When I am trying to use SSH key pair instead of the OS Login and IAP, it seems that the plugin attempts to connect to the internal IP address (which is obviously unaccessible for me, since I'm not running packer in GCP) and times out with:

>==> googlecompute.basic-example: Waiting for SSH to become available...
>==> googlecompute.basic-example: Timeout waiting for SSH.

tanmay-hc:
Introduces an option, `oslogin_ssh_username`, to customize the SSH username used with OS Login, enabling explicit control or normalization for federated identities. Improves compatibility across authentication methods and documents the new setting for users.

The option supports the following sentinel values:
- `__auto__` [default]: Fetches the username from the OS Login profile from GCP.
- `__external__`: Adds an `ext_` to the login profile username and truncates it to 32 chars.

Closes #254

### DIFF
--- a/builder/googlecompute/step_import_os_login_ssh_key.go
+++ b/builder/googlecompute/step_import_os_login_ssh_key.go
@@ -20,6 +20,11 @@ import (
 	"google.golang.org/api/oauth2/v2"
 )
 
+const (
+	AutoOSLoginUser     = "__auto__"
+	ExternalOSLoginUser = "__external__"
+)
+
 // StepImportOSLoginSSHKey imports a temporary SSH key pair into a GCE login profile.
 type StepImportOSLoginSSHKey struct {
 	Debug         bool
@@ -120,6 +125,25 @@ func (s *StepImportOSLoginSSHKey) Run(ctx context.Context, state multistep.State
 			username = account.Username
 			break
 		}
+	}
+
+	// The SSH keys uses the `loginProfile` username
+	config.loginProfileUsername = username
+
+	switch config.OSLoginSSHUsername {
+	case AutoOSLoginUser, "":
+		ui.Say("Using SSH Username for OSLogin...")
+	case ExternalOSLoginUser:
+		ui.Say("Using External SSH Username for OSLogin...")
+		if config.OSLoginSSHUsername == ExternalOSLoginUser && !strings.HasPrefix(username, "sa_") && !strings.HasPrefix(username, "ext_") {
+			username = "ext_" + username
+		}
+	default:
+		username = config.OSLoginSSHUsername
+	}
+
+	if len(username) > 32 {
+		username = username[:32]
 	}
 
 	log.Printf("[DEBUG] OSLogin step, setting ssh_username: %s", username)
--- a/builder/googlecompute/step_import_os_login_ssh_key.go
+++ b/builder/googlecompute/step_import_os_login_ssh_key.go
@@ -130,24 +130,8 @@ func (s *StepImportOSLoginSSHKey) Run(ctx context.Context, state multistep.State
 	// The SSH keys uses the `loginProfile` username
 	config.loginProfileUsername = username
 
-	switch config.OSLoginSSHUsername {
-	case AutoOSLoginUser, "":
-		ui.Say("Using SSH Username for OSLogin...")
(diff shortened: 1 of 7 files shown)


builder/googlecompute/step_import_os_login_ssh_key.go
```diff
@@ -122,6 +127,29 @@ func (s *StepImportOSLoginSSHKey) Run(ctx context.Context, state multistep.State
...
+	config.loginProfileUsername = username
+
+	switch config.OSLoginSSHUsername {
+	case AutoOSLoginUser:
+		ui.Say("Using SSH Username for OSLogin...")
+	case ExternalOSLoginUser:
+		ui.Say("Using External SSH Username for OSLogin...")
+		if config.OSLoginSSHUsername == ExternalOSLoginUser && !strings.HasPrefix(username, "sa_") && !strings.HasPrefix(username, "ext_") {
+			username = "ext_" + username
+			if len(username) > 32 {
```
sheshibodla, line 140: Given this condition is not specific to this case, moving this to after the switch/case statement may be more readable

builder/googlecompute/step_import_os_login_ssh_key.go
```diff
@@ -122,6 +127,25 @@ func (s *StepImportOSLoginSSHKey) Run(ctx context.Context, state multistep.State
...
 
+	// The SSH keys uses the `loginProfile` username
+	config.loginProfileUsername = username
+
+	switch config.OSLoginSSHUsername {
+	case AutoOSLoginUser, "":
+		ui.Say("Using SSH Username for OSLogin...")
+	case ExternalOSLoginUser:
+		ui.Say("Using External SSH Username for OSLogin...")
+		if config.OSLoginSSHUsername == ExternalOSLoginUser && !strings.HasPrefix(username, "sa_") && !strings.HasPrefix(username, "ext_") {
```
anshulsharma-hashicorp, line 138: part of the if condition `config.OSLoginSSHUsername == ExternalOSLoginUser` is not required as the same thing getting checked in case. and please add unit test case for this switch case to cover the expected username in different cases.

Issue #173, opened 2023-07-06
Support max-run-duration and instance-termination-action


ralbertazzi:
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

We have automated Packer jobs running in our CI systems. Sometimes Packer encounters temporary non-reproducible errors that lead to VMs that are not shutdown properly. This has several downsides both in terms of costs and robustness of future jobs (as running VMs contribute to the project's quota).

We learnt to live with this flakiness and embraced the fact that things can sometimes fail. We now have a periodic cleanup job that deletes packer VMs that have been running for too long, as it means that the related job probably failed in an ungraceful way.

Google Cloud supports setting the maximum runtime of an instance via the `max-run-duration` and `instance-termination-action` parameters. If this Packer plugin supported setting these values through HCL language, we would be able to get rid of the custom cleanup job.

#### Use Case(s)

Any relevant use-cases that you see.

#### Potential configuration

```
max_run_duration
instance_termination_action
```

PR #285, opened 2025-06-17
FEAT: Added Instance MaxRunDuration and Termination action

tanmay-hc:
This change allows user to use the MaxRunDuration and InstanceTerminationAction config params to kill their instances after a certain period.

As described in the issue, due to some unexpected issues, there are cases when Packer could not complete the job and the instances don't get deleted. This flag would help the users in cleaning up the instances automatically without any extra effort.

Closes https://github.com/hashicorp/packer-plugin-googlecompute/issues/173


kp2099 (approved):
lgtm

PR #324, opened 2025-12-28
add retry loop around ImportSshPublicKey

### LINKED ISSUE #320: Implement automatic retries for HTTP 409 errors from ImportOSLoginSSHKey
We're constantly running into issues where we start two completely unrelated builds simultaneously, and one of them fails with the following error:

```
googlecompute.sc-rocky-baseimage: Checking image does not exist...
googlecompute.sc-rocky-baseimage: Creating temporary RSA SSH key for instance...
googlecompute.sc-rocky-baseimage: no persistent disk to create
googlecompute.sc-rocky-baseimage: Importing SSH public key for OSLogin...
googlecompute.sc-rocky-baseimage: Error importing SSH public key for OSLogin: googleapi: Error 409: Multiple concurrent mutations were attempted. Please retry the request.
googlecompute.sc-rocky-baseimage: Deleting SSH public key for OSLogin...
googlecompute.sc-rocky-baseimage: SSH public key for OSLogin has been deleted!
googlecompute.sc-rocky-baseimage' errored after 3 seconds 304 milliseconds: Error importing SSH public key for OSLogin: googleapi: Error 409: Multiple concurrent mutations were attempted. Please retry the request.
```

The code where the error originates is here: https://docs.cloud.google.com/go/docs/reference/cloud.google.com/go/oslogin/latest/apiv1#cloud_google_com_go_oslogin_apiv1_Client_ImportSshPublicKey

It would be preferable if Packer could retry the operation in these cases, since retrying the build will almost certainly make the error "go away". As far as I can tell the Go library used to interface with Google Cloud Platform doesn't support any retry mechanism though, so I'm not confident in trying to implement this myself.

arnej27959:
### Description
This PR adds a simple retry loop for the case where several (possibly unrelated) packer jobs run in the same account/project triggering a simultaneous add of "oslogin" ssh keys.  We recently hit this as we build several packer images for GCE, and found that https://github.com/hashicorp/packer-plugin-googlecompute/issues/320 described the issue already.  Since we already program in go a retry loop was simple to add; the error message helpfully tells us explicitly that trying again is a valid strategy.  Since the file already imports `crypto/rand` I used that for getting a random interval between retries.

### Resolved Issues

Closes #320

### Rollback Plan

If a change needs to be reverted, we will roll out an update to the code within 7 days.

### Changes to Security Controls

No changes to Security Controls.

### DIFF
--- a/lib/common/driver_gce.go
+++ b/lib/common/driver_gce.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	compute "google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
 	impersonate "google.golang.org/api/impersonate"
 	oauth2_svc "google.golang.org/api/oauth2/v2"
 	"google.golang.org/api/option"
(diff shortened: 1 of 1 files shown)


tanmay-hc (changes requested):
Thank you for the PR @arnej27959 . This looks like a good fix.

I just have one concern regarding the error handling, relying on `strings.Contains(err.Error(), ...)`. This seems brittle. The API error messages can change without warning. Can we instead use the following snippet?

```
const maxRetries = 10

for i := 0; i < maxRetries; i++ {
    resp, err := d.osLoginService.Users.ImportSshPublicKey(parent, sshKey).Do()
    if err == nil {
        return resp.LoginProfile, nil
    }

    if gErr, ok := err.(*googleapi.Error); ok && gErr.Code == 409 {
        log.Printf("ImportSshPublicKey conflict (try %d/%d): %v", i+1, maxRetries, err)
        sleepSeconds := 5 + rand.Intn(11) 
        time.Sleep(time.Duration(sleepSeconds) * time.Second)
        continue
    }
    return nil, err
}
```

This seems to simplify the loop as well. Let me know your thoughts.

PR #348, opened 2026-06-08
fix: support universe regions starting with u- in zoneRegexp

jnahelou:
## Description
This PR updates the zone regular expression (`zoneRegexp`) to correctly support universe regions (which follow the format `u-<country>-<direction><number>-[a-z]`, e.g., `u-france-east1-a`). 

Without this fix, the regex fails to match zones with multiple hyphens before the number, preventing the successful deletion of disks during the cleanup phase.

This ensures both standard regions (europe-west9-a) and universe regions (u-france-east1-a) are properly matched and parsed.

## Related Logs / Error Output
When attempting to delete disks in these regions, the process fails with the following error:

```
==> googlecompute.debian_custom: DiskName: packer-uid
==> googlecompute.debian_custom: Zone: u-france-east1-a
==> googlecompute.debian_custom: Error: googleapi: Error 400: Invalid value for field 'region': 'u-france-east1-a'. Unknown region., invalid
```

### DIFF
--- a/lib/common/block_device.go
+++ b/lib/common/block_device.go
@@ -267,7 +267,7 @@ func GetRegionFromZone(zone string) (string, error) {
 	return matches[1], nil
 }
 
-var zoneRegexp = regexp.MustCompile("^[a-z]+-[a-z]+[0-9]{1,2}-[a-z]$")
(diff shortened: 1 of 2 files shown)


jnahelou:
Please let me know if you need any changes or additional details.
Thanks!

tanmay-hc (approved):
LGTM. Thanks for the PR!

PR #363, opened 2026-09-21
BUG: Fix panic when setting image deprecation status fails

### LINKED ISSUE #362: googlecompute: panic 'interface conversion: string is not error' at builder.go:131 after images.insert succeeds (v1.2.7)
### Versions

- packer `1.15.4`
- packer-plugin-googlecompute `1.2.7` (still the latest published release as of 2026-09-18)

### What happens

A plain `googlecompute` builder — standard Ubuntu 24.04 `noble` source image, a handful of shell
provisioners, image labels set — panics intermittently in the plugin's post-`images.insert`
wait/retry path, **after** the image has already been created successfully:

```
Build '<build-name>' errored after 8 minutes 32 seconds: unexpected EOF
--> <build-name>: unexpected EOF
==> Builds finished but no artifacts were created.
panic: interface conversion: string is not error: missing method Error
<date> packer-plugin-googlecompute_v1.2.7_x5.0_linux_amd64 plugin: goroutine 34 [running]:
<date> packer-plugin-googlecompute_v1.2.7_x5.0_linux_amd64 plugin: 	github.com/hashicorp/packer-plugin-googlecompute/builder/googlecompute/builder.go:131 +0x101c
<date> packer-plugin-googlecompute_v1.2.7_x5.0_linux_amd64 plugin: created by net/rpc.(*Server).ServeCodec in goroutine 1
<date> ui error: Build '<build-name>' errored after 8 minutes 32 seconds: unexpected EOF
<date> machine readable: <build-name>,error []string{"unexpected EOF"}
<date> ui error: --> <build-name>: unexpected EOF
==> Builds finished but no artifacts were created.
```

`packer` then exits non-zero and writes no `manifest.json` — i.e. a build that fully succeeded is
reported as a hard failure with no artifact recorded.

### What precedes it

Every time we've hit this, the GCE image is confirmed to already exist and be `READY`, with the
labels the build set, by the time the panic fires — `images.insert` completed successfully and the
plugin was already past all provisioners (this is the last stage of the build). A first occurrence
(2026-07-13) showed a **second** `images.insert` call ~70s after the first one's success, which
returned an `alreadyExists` error; the panic looks like the plugin mishandling that error value
(a `string`, not an `error`) somewhere in its insert-retry/wait polling, i.e. a slow/ambiguous poll
on the first insert operation triggers a redundant insert, and the redundant insert's error response
isn't type-asserted safely.

### Impact

A fully successful, several-minutes-long image build is reported as a failed CI job with no
artifact — the automation built on top of `packer build`'s exit code / manifest has no way to tell
"real failure" from "succeeded, then panicked" apart from independently querying GCE for the image.

### Reproduction

We can't reproduce on demand — it's intermittent (twice observed so far, ~2 months apart, both on
v1.2.7) — but the shape is a minimal `googlecompute` source block with `image_labels` set, building
from an official `ubuntu-os-cloud` source image, invoked from an internal CI system (an ephemeral,
short-lived GCE builder VM per build). No custom retry/timeout overrides on the builder.

### Our mitigation (workaround, not a fix)

We now treat a non-zero `packer build` exit as inconclusive rather than "no image was created":
after a non-zero exit we call `gcloud compute images describe <derived-image-name>` and, if the
image exists, is `READY`, and its labels match what the build would have set, we treat the build as
succeeded and continue. That papers over the panic; we'd rather not need it.

Happy to share full logs if useful (redacted for an internal/private CI system).

tanmay-hc:
## Problem

`Builder.Run` type-asserts the state bag's `"error"` value to `error`:

```go
return nil, rawErr.(error)
```

Two steps stored a `string` there instead. When `images.deprecate` failed, the plugin panicked with `interface conversion: string is not error` — after the image had already been created successfully. Packer surfaces this only as `unexpected EOF` with no artifact, so a green build is reported as a hard failure.

Two things made it intermittent rather than constant:

- `images.deprecate` is the last API call of the build, so it needs an otherwise perfect ~8 minute run to be reached.
- It's a single unretried HTTP request (`gensupport.SendRequest` → one `client.Do`), unlike the preceding operation poll, which retries indefinitely. One dropped connection was fatal.

Every other error path stores a real `error` and fails cleanly, so only this one call turned a routine blip into a process crash.

## Behavior

No user-visible change. The skipped call was already a no-op, and builds that do configure `deprecate_at` / `obsolete_at` / `delete_at` behave as before, now with retries.

Fixes #362


lib/common/driver_gce.go
```diff
@@ -285,8 +286,54 @@ func (d *driverGCE) SetImageDeprecationStatus(project, name string, deprecationS
 	if deprecationStatus == nil {
 		return errors.New("deprecationStatus cannot be nil")
 	}
-	_, err := d.service.Images.Deprecate(project, name, deprecationStatus).Do()
-	return err
+
+	// Images.Deprecate is a single unretried HTTP request, and it runs at the
+	// very end of a build, so a transient failure here throws away an image
+	// that was already created successfully. Retry the transient cases.
+	return retry.Config{
```
taru-garg-2000: [nit]: I would recommend storing the error, currently it given an impression we are returning a `retry.Config`

taru-garg-2000 (commented):
Left 1 nit but looks good

lib/common/driver_gce.go
tanmay-hc: Its a `retry.Config{}.Run(...return error)`. I would still rewrite this.

taru-garg-2000 (approved):

